PR #5, opened 2022-04-28
Max branch

TeachRaccooon:
Existing code: added templates to utils namespace functions.
New code: added dense operator generation + testing.
Included some comments with explicit questions in the beginning of implementation and testing files.

### DIFF
--- a/src/dense_op.cc
+++ b/src/dense_op.cc
@@ -17,15 +17,10 @@ namespace RandBLAS::dense_op {
 /*
 Note from Random123: Simply incrementing the counter (or key) is effectively indistinguishable
 from a sequence of samples of a uniformly distributed random variable.
-Question: does it matter in which order (row vs col) we fill in data here?
 
 Notes:
-1. Is using several templates here a bad decision? Currently, Not sure how to make generalized implementations
-   to support both 2 and 4 random number generations.
-2. With every function run, at least 1 and up to 3 generated random numbers will not be used. Is it worth storing those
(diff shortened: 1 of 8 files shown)


rileyjmurray (commented):
Great work! I left a few comments, mostly in ``dense_op.cc``. Let's discuss those before looping in Burlen for a proper review.

src/dense_op.cc
```diff
@@ -0,0 +1,190 @@
...
+2. Note on eventually needing to add 8 and 16-bit generations.
+*/
+
+
+
+// Actual work - uniform dirtibution
+template <typename T, typename T_gen>
+static void gen_unif(int64_t n_rows, int64_t n_cols, T* mat, uint32_t seed)
+{
+        typename T_gen::key_type key = {{seed}};
```
rileyjmurray, line 32: A couple things: I think we should take in a pointer to a uint32_t as a seed, use it as a counter (rather than a key), and increment it as we go along. It would also be good for us to accept a key separately from the seed. That way we could support embarrassingly parallel computations. The semantics would always be that "seed" is passed as a pointer and incremented while "key" is not.

rileyjmurray (approved):
There are some changes I'd like to make in the API and code style, but I don't want those to complicate your QB PR for RandLAPACK. Once we merge that PR I can go back and fix RandBLAS and update RandLAPACK (that gives me a chance to touch the code a little even if I can't work on it really actively right now).

PR #16, opened 2022-12-01
Viable API for dense and sparse sketching

rileyjmurray:
This PR contains a complete rewrite of RandBLAS' API. The important functionality is declared in three headers: ``base.hh``, ``sparse.hh``, and ``dense.hh``, which define functionality in the ``RandBLAS::base``, ``RandBLAS::sparse``, and ``RandBLAS::dense`` namespaces. Here's a breakdown of what can be found in each of these headers

## RandBLAS/include/base.hh

This defines functionality for the counter-based random number generators (CBRNGs) that RandBLAS uses for basic random number generation.

The state of a CBRNG is specified by a "counter" and a "key". Conceptually these can be thought of as 128-bit or 256-bit integers. The key specifies the stream of numbers that a CBRNG will produce. One accesses random numbers within the stream by specifying the value of the counter.

From an implementation standpoint, there is a bit of complexity in representing counters and keys. This is because the 128-bit / 256-bit integers will actually be represented by arrays of simpler numeric types. The functionality defined in this header file tries to hide that complexity. It also hides the fact that we're relying on Random123 for implementation of our CBRNGs.

### Important definitions

``RNGName`` is a character-based enumeration used to specify the type of CBRNG.
* ``RNGName::Philox`` is associated with the``RandBLAS::base::Philox`` CBRNG (which is just an alias of ``r123::Philox4x32``.
* ``RNGName::Threefry`` is associated with the ``RandBLAS::blase::Threefry`` CBRNG (which is just an alias of ``r123::Threefry4x32``).

``RNGState`` is a type of struct for carrying counters, keys, and an RNGName.
* These structs appear in the public API.
* Users should construct ``RNGState`` structs by specifying a 32-bit or 64-bit unsigned integer for the counter and a 32-bit unsigned integer for the key.

``Random123_RNGState<T_gen>`` is a type of struct that carries counters and keys for the CBRNG ``T_gen``.
* The counters and keys are not "just" arrays of unsigned 32-bit ints; they're Random123-specific data structures. 
* This is an implementation artifact and *not* part of the public API.

## RandBLAS/include/sparse.hh

This defines the API for constructing and using sparse sketching operators. The function ``lskges`` reads as "left-sketch a general matrix, with a sparse sketching operator". Important definitions are below.

``SparseDistName`` is a character-based enumeration that specifies qualitative information in a sparse sketching operator. It's used to indicate whether one wants to work with "short-axis-sparse operators" (SASOs) or "long-axis-sparse operators" (LASOs), in the sense of the RandLAPACK book. Right now, ``sparse.cc`` only implements SASOs, and the SASOs must have more columns than rows. Note that any sketching operator with more columns than rows can only be used to sketch from the left.

``SparseDist`` is a type of struct that specifies the distribution over sparse sketching operators. ``SparseDist.family`` specifies whether we're working with SASOs or LASOs, while ``SparseDist.n_rows`` and ``SparseDist.n_col`` specify the dimensions of the sketching operator. ``SparseDist.vec_nnz`` is the number of nonzeros in each short-axis vector (for SASOs) or long-axis vector (for LASOs).

``SparseSkOp`` is a type of struct for representing a specific sparse sketching operator. The distribution from which the sketching operator is sampled is specified in ``SparseSkOp.dist``. The state of the CBRNG used to define the sketching operator is given in ``SparseSkOp.seed_state``. After a program defines a ``SparseSkOp`` object ``S``, the next time it needs to seed a CBRNG it should do so with ``S.next_state``.

## RandBLAS/include/dense.hh

This defines the API for constructing and using dense sketching operators. The function ``lskge3`` reads as "left-sketch a general matrix, with a BLAS 3 operation". This header file defines similar enumerations and structs as ``sparse.hh``. Differences are highlighted below.

``DenseDistName`` is a character-based enumeration analogous to ``SparseDistName``. The options are  ...
* Gaussian,
* Uniform (uniform over [-1, 1]),
* Rademacher (uniform over {+1, -1}),
* Haar (uniform over row-orthonormal or column-orthonormal matrices), and
* DisjointIntervals (uniform over $[-2, -1] \cup [1, 2]$).

Of these, only Gaussian and Uniform are implemented.

``DenseDist`` is a type of struct that's analogous to `` SparseDist``. The only difference between these two structs is that the former lacks the ``.vec_nnz`` field of the latter.

``DenseSkOp`` is a type of struct that's analogous to ``SparseSkOp``. One difference is that it features an attribute ``DenseDist.layout`` which says whether the buffer that represents a sketching operator should be interpreted in row-major or column-major order. Implementations in ``dense.cc`` also allow lazy instantiation of the memory that's needed for a ``DenseSkOp`` object. The behavior of lazy memory allocation or lazy random number generation are governed by ``DenseSkOp.filled``, ``DenseSkOp.persistent``, and ``DenseSkOp.buff``.


src/dense.cc
```diff
@@ -0,0 +1,295 @@
...
+
+static inline void sincospif(float x, float *s, float *c) {
+    const float PIf = 3.1415926535897932f;
+    sincosf(PIf*x, s, c);
+}
+
+static inline void sincospi(double x, double *s, double *c) {
+    const double PI = 3.1415926535897932;
+    sincos(PI*x, s, c);
+}
```
rileyjmurray, line 44: This is to handle compilation issues that can arise when compiling on macOS and/or compiling with CUDA.

src/sparse.cc
```diff
@@ -0,0 +1,345 @@
+#include "sparse.hh"
```
rileyjmurray, line 1: The indentation throughout this file is pretty messed up. Needs a full pass.

test/src/test_dense.cc
```diff
@@ -0,0 +1,443 @@
...
+                D, seed, 0, buff.data(), false, true, layout
+            );
+        } else {
+            S0_ptr = new RandBLAS::dense::DenseSkOp<T>(
+                D, seed, 0, buff.data(), false, true, layout
+            );
+        }
+
+        // define a matrix to be sketched, and create workspace for sketch.
+         bool is_colmajor = layout == blas::Layout::ColMajor;
```
rileyjmurray, line 129: Indentation error

test/src/test_dense.cc
```diff
@@ -0,0 +1,443 @@
...
+        // Define the distribution for S0.
+        RandBLAS::dense::DenseDist Dt = {
+            .family=RandBLAS::dense::DenseDistName::Gaussian,
+            .n_rows = m,
+            .n_cols = d
+        };
+        // Define the sketching operator struct, S0.
+        RandBLAS::dense::DenseSkOp<T> S0(Dt, seed, 0,
+            NULL, false, true, layout
+        );
```
rileyjmurray, line 167: Default arguments are great. However, whether or not some default arguments are left to their default values depends heavily on context. The majority of users wouldn't care about setting ``layout``, since it defaults to ``ColMajor``. But if the user wants to specify "use a RowMajor layout, but only once it becomes necessary to allocate memory", then they have to provide the explicit defaults (here, ``NULL, false, true``, in order).

If we move ``layout`` earlier in the argument list then we run into the same problem with other arguments. For example, we usually have ``persistent=true``, but we might want ``persistent=false`` and to use column-major layout without explicitly stating as much. This is kind of a red herring though since I might be able to do away with ``layout`` in the ``DenseSkOp`` class in light of moving from "pointer offsets" to ``(i_op, j_op)``.

Here's a simpler example. At time of leaving this comment, the following function call works:
```
RandBLAS::sparse::SparseSkOp<T> sas(
    RandBLAS::sparse::SparseDistName::SASO,
    n_rows, n_cols, vec_nnz, key, ctr_offset
);
```
However, the following instantiation (of what seems to be an equivalent struct) fails
```
    RandBLAS::sparse::SparseSkOp<T> sas = {
        .family = RandBLAS::sparse::SparseDistName::SASO,
        .n_rows = n_rows,
        .n_cols = n_cols,
        .vec_nnz = vec_nnz,
        .key = key,
        .ctr_offset = ctr_offset
    };
```

weslleyspereira: Hi. My suggestion is not bother too much with the order of the parameters. This may be difficult to maintain in a long term since your struct has lots of arguments. I you provide different kinds of constructors, which will have their own default arguments. Also, since you have a struct, you can always set values after the object is created:

```c++
RandBLAS::sparse::SparseSkOp<T> sas( ... );
sas.persistent = false;
```

include/RandBLAS/dense.hh
```diff
@@ -0,0 +1,181 @@
...
+
+    DenseSkOp(
+        DenseDist dist,
+        uint32_t ctr_offset,
+        uint32_t key,
+        T *buff,
+        bool filled,
+        bool persistent,
+        blas::Layout layout
+    ) : DenseSkOp(dist, RNGState{ctr_offset, key}, buff, filled, persistent, layout) {};
```
rileyjmurray, line 85: Note to self: It seemed that for SparseSkOp to work, I had to move these kinds of constructors into the .cc file, not leave in the .hh file. But this suggests leaving in the .hh file is fine.

PR #24, opened 2022-12-19
Refactor header only

burlen:
this restructures directories to a more modern style and moves template implementations into header files. we are now very close to a header only library, as only base.cc requires compilation because of RNGState. Templating RNGState will simplify things and would get us to header only, however, if that happens it will be in another PR. This also cleans up the various Random123 work arounds.


RandBLAS/dense.hh
```diff
@@ -1,27 +1,189 @@
...
+
+
+
 template <typename T, typename T_gen>
 static RNGState gen_unif(
     int64_t n_rows,
     int64_t n_cols,
     T* mat,
-    RNGState state
+    const RNGState &state
```
rileyjmurray, line 186: Originally I passed by value so that it didn't matter if we made changes to ``RNGState``. My guess is that this change of passing by a const reference is slightly more efficient. Is that reading correct?

RandBLAS/dense.hh
```diff
@@ -97,29 +255,23 @@ static RNGState gen_norm(
     rout = gen(impl_state.ctr, impl_state.key);
     pair_1 = r123::boxmuller(rout.v[0], rout.v[1]);
     pair_2 = r123::boxmuller(rout.v[2], rout.v[3]);
-    T *v = new T[4] {pair_1.x, pair_1.y, pair_2.x, pair_2.y};
+    T v[4] = {pair_1.x, pair_1.y, pair_2.x, pair_2.y};
```
rileyjmurray, line 258: What's the significance of using this syntax relative to the old syntax?

rileyjmurray: Is it just that the compiler will automatically handle deleting ``v`` on-exit from the function?

RandBLAS/random_gen.hh
```diff
@@ -0,0 +1,20 @@
...
+#include <math.h>
+#include <Random123/array.h>
+#include <Random123/philox.h>
+#include <Random123/uniform.hpp>
+#include <Random123/philox.h>
+#include <Random123/threefry.h>
+#include <Random123/boxmuller.hpp>
+
+using Philox = r123::Philox4x32;
+using Threefry = r123::Threefry4x32;
```
rileyjmurray, line 18: Earlier I used typedefs to create aliases for these types. How does this ``using`` assignment compare to a typedef? From a source-code perspective, I'm guessing there's no real difference. Does it result in less-messy compiler errors?

RandBLAS/util.hh
```diff
@@ -1,11 +1,15 @@
...
-#include <string>
-#include <stdexcept>
-#include <stdio.h>
+#ifndef randblas_util_hh
+#define randblas_util_hh
+
+#include <blas.hh>
+#include <cstdio>
+//#include <iostream>
+//#include <string>
```
rileyjmurray, line 7: These lines can probably be removed.

RandBLAS/sparse.hh
```diff
@@ -140,14 +251,16 @@ static void sketch_cscrow(
     int64_t ldb,
     int threads
```
rileyjmurray, line 252: This used to have a default value of 4. Can we add that back in? I'd like for ``lskges`` to have the same signature as ``lskge3``, and the latter doesn't have a required ``threads`` argument.

test/test_sparse.cc
```diff
@@ -162,7 +162,7 @@ class TestLSKGES : public ::testing::Test
                 layout, blas::Op::NoTrans, blas::Op::NoTrans,
                 d, n, m,
                 1.0, sas, 0, 0, a, lda,
-                0.0, a_hat, ldahat
+                0.0, a_hat, ldahat, 1
```
rileyjmurray, line 165: Related to my comment in the new ``RandBLAS/sparse.hh`` file: ``lskges`` used to have a default value of 4 for this parameter. I'd like to go back and continue to use a default value for this parameter so that ``lskges`` and ``lskge3`` can have the same signature.

rileyjmurray (commented):
Mostly looks great! I've asked a few questions and made some minor requests. Please address when you get a chance.

include/RandBLAS/base.hh
```diff
@@ -1,123 +0,0 @@
...
-    static inline void sincospif(float x, float *s, float *c) {
-        const float PIf = 3.1415926535897932f;
-        sincosf(PIf*x, s, c);
-    }
-
-    static inline void sincospi(double x, double *s, double *c) {
-        const double PI = 3.1415926535897932;
-        sincos(PI*x, s, c);
-    }
-#endif
```
rileyjmurray, line 105: @burlen how confident are you that the change in ``RandBLAS/random_gen.hh`` captures these workarounds? While I know you've experienced the issue that led to the ``!defined(__CUDACC__)`` check, I made to personally add the ``#if !defined(R123_NO_SINCOS) && defined(__APPLE__)`` check.

RandBLAS/dense.hh
burlen: that makes sense. passing by reference to const is more efficient because we skip the deep copy made by the copy constructor. because it is a ref to const compiler will ensure we do not modify it. my 2 cents  : when passing an object by value intentionally we should add a comment that we're doing it on purpose because it is usually a mistake to pass by value and someone like me might change it leading to a bug.

RandBLAS/dense.hh
burlen: The `new  T[4]` allocates memory on the heap which is relatively expensive operation. The same goes for `delete []`. We want to avoid using those when we can. `T v[4]` allocates space on the stack at compile time which is much faster. It's best to put small fixed size arrays on the stack. There are limits to how much you can put on the stack (too much leads to a "stack overflow" ;)). A 4 element array is safe.

RandBLAS/random_gen.hh
burlen: A typedef and an alias are different features. Aliases can be templated while typedefs cannot.  Otherwise they are equivalent. my opinion is that as part of our code style we might use aliases every where for consistency sake. a minor issue though.

RandBLAS/sparse.hh
burlen: Sure. I will add it back in. What  are your thoughts on making the default 1? 

For instance a default of N > 1 threads will not yield good performance if there are M < N physical cores present. On the other hand if you go with OpenMP's default then it allocates a thread for each logical core (hyperthread). this is usually bad for performance. the best performance for numerical codes is almost always 1 thread per physical core. When MPI is in play (thinking about the future) OpenMP thread pools should be sized such that the threads in each MPI process are uniquely mapped to physical cores across a node. With P MPI ranks per node, and M physical cores per node, a default of N > 1 OpenMP threads will result in poor performance when N * P > M.

burlen: The usual way to specify number of threads is the `OMP_NUM_THREADS` and `OMP_PLACES` environment variables. `omp_set_num_threads` documentation overrides indicates that it overrides those controls, so I think we should avoid using it, or at the very least check in the environment for `OMP_NUM_THREADS` and only call it if it is not set.

burlen: It's probably best if we never call `omp_set_num_threads` internally in the library and leave it up to the code calling the library to do it right. this is where documentation becomes critical.

test/test_sparse.cc
burlen: OK, agree with you about them having the same signature.

include/RandBLAS/base.hh
burlen: I am confident that on Linux we don't need it. I added the following to `random_gen.h`
```
// this is for sincosf
#ifndef _GNU_SOURCE
#define _GNU_SOURCE
#endif
```
after noticing from the manpage for `sincosf` that `_GNU_SOURCE` needed to be defined to get them. Tested this on both Fedora and Ubuntu. 

Untested on MacOS. Let me check that before merging this.

burlen: The function is named `__sincosf` on MacOS. My change probably breaks on MacOS, the work around may be to make `#define sincosf __sincosf` on MacOS.

RandBLAS/dense.hh
rileyjmurray: Sounds good; I'm on board with explicitly acknowledging passing by value whenever we do need it (which I imagine will be rare).

RandBLAS/random_gen.hh
rileyjmurray: Gotcha. I just looked into aliases vs typedefs, and I agree with your conclusion.

RandBLAS/sparse.hh
rileyjmurray: If you say it's best that we never call ``omp_set_num_threads`` in the library, that's fine. Can the user call ``omp_set_num_threads`` before and after a call to a RandBLAS function they want to benchmark? That is, can the user do something like the following?
```
omp_set_num_threads(4);
lskges(....);
omp_set_num_threads(... default ...);
```

include/RandBLAS/base.hh
rileyjmurray: I'm on board with using ``#define sincosf __sincosf`` on MacOS.

RandBLAS/sparse.hh
rileyjmurray: Seems the answer to my question is yes: https://stackoverflow.com/questions/22813923/difference-between-num-threads-vs-omp-set-num-threads-vs-omp-num-threads.

include/RandBLAS/base.hh
rileyjmurray: Hmm. I just tried compiling the code on this branch (which does not have ``#define sincosf __sincosf`` at time of writing) and it worked just fine on my mac. I guess maybe the block
```
// this is for sincosf
#ifndef _GNU_SOURCE
#define _GNU_SOURCE
#endif
```
worked for macOS and linux? In any case, I think it's safe to not make any more changes to the Random123 workaround.

PR #25, opened 2022-12-21
WIP -- Rng types

burlen:
some experiments w/ random123. These have been included into #28. Note: in #28 I squashed the commits marked squash here.

PR #28, opened 2022-12-27
integrate templated RNGState

burlen:
This contains everything on #25 including prototypes and improvements to the dense implementation as well as integration for sparse. On the sparse side I tried to make minimal changes.


RandBLAS/base.hh
```diff
@@ -1,131 +1,191 @@
...
-        this->rng_name = RNGName::Philox;
-    } else if (gtid == ::std::type_index(typeid(tf))) {
-        this->rng_name = RNGName::Threefry;
-    } else {
-        throw std::runtime_error(std::string("Unknown gen_type."));
-    }
+template <typename RNG>
+RNGState<RNG>::operator std::tuple<typename RNG::ctr_type&, typename RNG::key_type&> () {
+    return std::tie(counter, key);
+}
```
rileyjmurray, line 86: What are these functions supposed to do?

RandBLAS/random_gen.hh
```diff
@@ -9,12 +9,28 @@
...
 #include <Random123/boxmuller.hpp>
+#include <Random123/uniform.hpp>
+
+/// @cond
+/// extend r123::float2 to work with structured bindings
+namespace std {
+  template<> struct tuple_size<r123::float2> { static constexpr size_t value = 2; };
+  template<> struct tuple_element<0, r123::float2> { using type = float; };
+  template<> struct tuple_element<1, r123::float2> { using type = float; };
+}
```
rileyjmurray, line 24: Is modifying the std namespace like this considered common practice?

RandBLAS/sparse.hh
```diff
@@ -136,18 +136,18 @@ SparseSkOp<T>::SparseSkOp(
...
         delete [] this->cols;
         delete [] this->vals;
     }
 };
 
-template <typename T>
+template <typename SKOP>
 static bool has_fixed_nnz_per_col(
-    SparseSkOp<T> &S0
+    SKOP const& S0
```
rileyjmurray, line 150: Why is this preferable to ``SparseSkOp<T>``?

RandBLAS/sparse.hh
```diff
@@ -170,29 +171,26 @@ static RNGState repeated_fisher_yates(
     for (int64_t j = 0; j < vec_len; ++j)
         vec_work[j] = j;
     std::vector<int64_t> pivots(vec_nnz);
-    typedef typename T_gen::ctr_type ctr_type;
-    ctr_type rout;
-    T_gen g;
-    RNGState out_state(init_state);
+    RNG gen;
+    auto [ctr, key] = state;
```
rileyjmurray, line 175: What makes this kind of unpacking possible? Is it the lines in https://github.com/BallisticLA/RandBLAS/pull/28/files#diff-8025ac5321d9ef206e31f5acca1b95378b5664b26ad732dcd4c4c66349eb3ce8R78-R86?

RandBLAS/sparse.hh
```diff
@@ -170,29 +171,26 @@ static RNGState repeated_fisher_yates(
...
-        Random123_RNGState<T_gen> impl_state(init_state);
-        impl_state.ctr.incr(offset);
+        auto ctri = ctr;
+        ctri.incr(offset);
         for (int64_t j = 0; j < vec_nnz; ++j) {
             // one step of Fisher-Yates shuffling
-            rout = g(impl_state.ctr, impl_state.key);
-            int64_t ell = j + rout.v[0] % (vec_len - j);
+            auto rv = gen(ctri, key);
+            int64_t ell = j + rv[0] % (vec_len - j);
```
rileyjmurray, line 183: Did you write https://github.com/BallisticLA/RandBLAS/pull/28/files#diff-38df2d18a53b7af91f2aa92f6dc8b40a0582eae533acb874f8ba3ba92681cc6fR26-R33 so you could index into ``rv`` like this? Is ``gen`` still a Random123 RNG?

RandBLAS/sparse.hh
```diff
@@ -234,32 +233,15 @@ static RNGState template_fill_sparse(
...
-        default:
-            throw std::runtime_error(std::string("Unrecognized generator."));
-    }
-}
-
-
-template <typename T>
-void print_saso(SparseSkOp<T>& S0) {
+template <typename SKOP>
+void print_saso(SKOP const& S0) {
```
rileyjmurray, line 244: Same question as before: why not just use ``SparseSkOp<T>``?

RandBLAS/sparse.hh
```diff
@@ -578,7 +559,7 @@ void lskges(
     int64_t n, // op(A) is m-by-n
     int64_t m, // op(S) is d-by-m
     T alpha,
-    SparseSkOp<T> &S0,
+    SparseSkOp<T,RNG> &S0,
```
rileyjmurray, line 562: I have a strong preference for the RNG not to appear in this function signature. How can we make sure to present an API to the users that doesn't need this?

RandBLAS/base.hh
```diff
@@ -1,131 +1,191 @@
...
+        <sizeof(typename RNG::ctr_type::value_type) == sizeof(uint32_t), float, double>::type>
+    static
+    auto generate(
+        RNG &rng,
+        typename RNG::ctr_type const& c,
+        typename RNG::key_type const& k
+    ) {
+        return r123::uneg11all<T>(rng(c,k));
+    }
+};
```
rileyjmurray, line 188: My plan was for ``RandBLAS/base.hh`` to not include methods for sampling from different distributions. I'd like for us to specify clear semantics for ``RandBLAS/base.hh`` versus ``RandBLAS/random_gen.hh``.

test/test_sparse.cc
```diff
@@ -428,7 +428,7 @@ class TestLSKGES : public ::testing::Test
         uint32_t ctr_A0 = 42;
         uint32_t seed_A0 = 42000;
         RandBLAS::dense::DenseDist DA0 = {.n_rows = m0, .n_cols = n0};
-        RandBLAS::dense::fill_buff(A0.data(), DA0, RandBLAS::base::RNGState{ctr_A0, seed_A0});
+        RandBLAS::dense::fill_buff(A0.data(), DA0, RandBLAS::base::RNGState{{{ctr_A0}}, {{seed_A0}}});
```
rileyjmurray, line 431: To me, this looks much harder to read. Is it necessary based on the changes you made to RNGState? If so, what changes lead to this new initialization syntax?

RandBLAS/dense.hh
```diff
@@ -68,33 +68,29 @@ struct DenseDist {
...
+    bool filled = false;             ///< a flag that indicates if the memory was initialized
+    bool persistent = true;          ///< ???
+
+    const blas::Layout layout = blas::Layout::ColMajor; ///< matrix storage order
 
-    /////////////////////////////////////////////////////////////////////
-    //
-    //      Member functions must directly relate to memory management.
-    //
-    /////////////////////////////////////////////////////////////////////
```
rileyjmurray, line 92: We should have prominent comments to differentiate between properties specific to dense sketching operators and properties that exist for any sketching operator. I would like the recurring design principles to be documented, if only briefly, wherever they appear.

RandBLAS/dense.hh
```diff
@@ -68,33 +68,29 @@ struct DenseDist {
...
 
-    T *buff = nullptr;
-    bool filled = false;
-    bool persistent = true;
-    const blas::Layout layout = blas::Layout::ColMajor;
+    using generator = RNG;
+    using state_type = RNGState<RNG>;
+    using buffer_type = T;
```
rileyjmurray, line 76: Why do you need to define ``using buffer_type = T;``? My guess is so that we can template w.r.t ``typename SKOP`` and then in implementations that work with a sketching operator ``S`` we access the numerical type with ``using T = S::buffer_type;``. Is that correct?

RandBLAS/dense.hh
```diff
@@ -177,126 +173,103 @@ DenseSkOp<T>::~DenseSkOp() {
...
+    {
+        // add the start index to the counter in order to make the sequence
+        // deterministic independent of the number of threads.
+        int ti = omp_get_thread_num();
+        int nt = omp_get_num_threads();
+
+        int64_t chs = nit / nt;
+        int64_t nlg = nit % nt;
+        int64_t i0 = chs * ti + (ti < nlg ? ti : nlg);
+        int64_t i1 = i0 + chs + (ti < nlg ? 1 : 0);
```
rileyjmurray, line 220: Can you give the full names behind these abbreviated variable names, via comments? Like:
```
        // chs = [word that starts with c] ... [other words] ...
        int64_t chs = nit / not;
        // nlg = [word that starts with l] ... [other words] ...
        int64_t nlg = nit % nt;
```

RandBLAS/dense.hh
```diff
@@ -177,126 +173,103 @@ DenseSkOp<T>::~DenseSkOp() {
 
 
 
-
-
-
-
-template <typename T, typename T_gen>
-static RNGState gen_unif(
+/** Fill a n by m matrix with random values. If RandBLAS is compiled with
```
rileyjmurray, line 176: Let's not say ``n \times m``; just say ``n_rows \times n_cols``.

RandBLAS/base.hh
```diff
@@ -1,131 +1,191 @@
...
+     */
+    template <typename RNG, typename T = typename std::conditional
+        <sizeof(typename RNG::ctr_type::value_type) == sizeof(uint32_t), float, double>::type>
+    static
+    auto generate(
+        RNG &rng,
+        typename RNG::ctr_type const& c,
+        typename RNG::key_type const& k
+    ) {
+        return r123::uneg11all<T>(rng(c,k));
```
rileyjmurray, line 186: I can't actually infer the return type here. My preference is for the return type to be explicit rather than ``auto``. If that's a big pain then I'd like to have some examples of return types in the docstring.

rileyjmurray (changes requested):
Thanks for all this work, @burlen! I don't have any big blocking requests, but there are several things that I don't really understand. It would be good if you could provide initial answers to these questions. I expect we'll need to have a longer conversation about your changes to ``RandBLAS/base.hh``, but I think that can wait.

RandBLAS/sparse.hh
rileyjmurray: I see in the unit tests that the generator needn't be specified explicitly. So I think we'll be fine and I just need to understand how the templating works, lol.

RandBLAS/base.hh
burlen: those are conversion functions so RNGState can be converted to a std::tuple, which is useful with structured bindings

RandBLAS/random_gen.hh
burlen: yes, from what I've read this is the only way.

RandBLAS/sparse.hh
burlen: it's simpler to use, fewer template parameter

RandBLAS/sparse.hh
burlen: this is because of the conversions to std::tuple you were asking about.

RandBLAS/random_gen.hh
burlen: here's a simple write up https://blog.tartanllama.xyz/structured-bindings/

RandBLAS/base.hh
burlen: update: I'll remove these since they're not used any ware, and structured bindings work out of the box with classes/structs that have all public non static member variables.

RandBLAS/random_gen.hh
burlen: update: I'll remove these since `r123::flkoat2` works out of the box with structured bindings due to its all public non-static member variables.

RandBLAS/sparse.hh
burlen: to simplify the usage, it's tedious to have to manually specify all of the nested types when the compiler can do it for us.

RandBLAS/sparse.hh
burlen: Agree with you on hiding RNG from the users as much as possible. As you've noticed, the user won't have to specify RNG if they're OK with our default selection, which is currently `Philox4x32`. They can swap `Philox4x32` out fairly easily if desired. It's also easy for us to change the default if we ever decide that `Philox4x32` is not what we want. 

However, one option to remove RNG here would be to template on the sketch operator instead of RNG. This would be similar to what I've done in `print_saso` above. I considered doing it that way and didn't come up with a strong reasons for or against.

RandBLAS/base.hh
burlen: OK. Good point. I will move those functions from base to random_gen

test/test_sparse.cc
burlen: Agree. The double braces disambiguate what we intend to initialize here. this can be simplified, for instance by adding constructors to `RNGState` that take single `uint32_t`'s. In going with this admittedly confusing syntax I was opting to reduce the code in `RNGState`. I'll push a simplification in a follow up

RandBLAS/dense.hh
burlen: yes, it's a way of getting the type of data stored, from with in templated code.

RandBLAS/dense.hh
burlen: lol, sorry about that. I'll fix this, here chs means chunk_size, it's a term used in the OMP spec

burlen: nlg is number of large chunks

RandBLAS/dense.hh
burlen: ok

RandBLAS/base.hh
burlen: it will be a messy template in this case. It is documented in the comment, but it may not be clear where n comes from. It's, the counter length, or 4 in Philox4x32. Let me see if I can make the comment easier to understand.

burlen:
@rileyjmurray thanks for the review. I see some improvements based on your input. I think can address your concerns. Will ping you when it's ready for another look.

RandBLAS/base.hh
burlen: resolved in c6077f65

RandBLAS/random_gen.hh
burlen: resolved in b5bdf79a

RandBLAS/base.hh
burlen: resolved in c9303c27

RandBLAS/dense.hh
burlen: resolved in 02977485

RandBLAS/dense.hh
burlen: resolved in 0220c604b

RandBLAS/base.hh
burlen: resolved in c9303c27bb08

RandBLAS/sparse.hh
burlen: resolved in 9dbc707a

test/test_sparse.cc
burlen: resolved in 6166587ca

RandBLAS/dense.hh
burlen: resolved in de486e0d

burlen:
@rileyjmurray I took a stab at addressing some of the issue you pointed out. Did I address your concerns? I'll hold off on merging this until I hear back from you.

RandBLAS/dense.hh
rileyjmurray: Amazing! God, I'm so glad you're part of this project. I knew this was an important thing that I wanted to do, but since I didn't know how to describe it, I didn't google it effectively.

RandBLAS/random_gen.hh
```diff
@@ -15,22 +17,103 @@
...
 #include <Random123/uniform.hpp>
 
-/// @cond
-/// extend r123::float2 to work with structured bindings
-namespace std {
-  template<> struct tuple_size<r123::float2> { static constexpr size_t value = 2; };
-  template<> struct tuple_element<0, r123::float2> { using type = float; };
-  template<> struct tuple_element<1, r123::float2> { using type = float; };
+/// our extensions to random123
+namespace r123ext
```
rileyjmurray, line 21: I _love_ this convention for organizing mild add-ons to Random123 functionality.

rileyjmurray (approved):
Looks wonderful!

RandBLAS/sparse.hh
rileyjmurray: Following up on this. I liked ``SparseSkOp<T> &S0`` because it clearly indicated that ``S0`` had to be a ``SparseSkOp``. Those structs had clearly specified semantics. By contrast, templating by ``SKOP`` makes the precise semantics of ``S0`` ambiguous.

Would you say this is a (more or less) unavoidable consequence of suppressing the presence of RNG from the templating? It's okay if that's the case; I just want to understand the constraints based on how C++ does templating. At the end of the day, the ambiguity of templating by ``SKOP`` can always be addressed with documentation that clearly explains what ``SKOP`` might be. But if you know of other ways to address this ambiguity I'd be interested in hearing about them.

rileyjmurray: It looks like the other way to resolve ambiguity would be to use "constraints" and/or "concepts" from [C++20](https://en.cppreference.com/w/cpp/language/constraints). I'd rather not move to C++20 just because of this. So I'll mark this conversation as resolved.

rileyjmurray:
@burlen I have no remaining questions. Please merge when you're ready!

PR #27, opened 2022-12-22
Complete implementation of LSKGES

rileyjmurray:
This PR supersedes PR #26. Big changes include:
* Added support for long-axis-sparse sketching operators, or "LASOs". Prior to this PR, we only supported short-axis-sparse operators, or "SASOs."
* Added support for transposes on the sketching operator and data matrix.
* Added support for "lifting" instead of just "sketching" with LSKGES and LSKGE3. More on this below.
* LSKGES previously had a helper function for sketching row-major data from the left with a (non-transposed) SASO. This PR removes that function. More on this below.
* Burlen's changes in PR #28 for improved templating of RNGState structs.

### Lifting versus sketching

We used to have a requirement that ``B = op(S) * op(A)`` had fewer rows than ``op(A)``. That requirement is valid if we only allow proper sketching. However, there are situations where we need to "lift" the result of a sketch back into the higher-dimensional space, in the sense that ``B = op(S) * op(A)`` has more rows than ``op(A)``.

### Removing specialized row-major sketching function

The new implementation for applying a CSC-like sketching operator accesses data with appropriate strides. I did this because it substantially simplified the case handling for SASOs, LASOs, transposes thereof, row-major and column-major data, and (eventually) right-sketching. However, accessing data via strides does introduce a performance penalty, so we should look at restoring the functionality I removed in the near future.

PR #31, opened 2023-01-06
testing add address sanitizer runs

burlen:
this will catch memory leaks, invalid memory read/write, etc...

let's get #27 in before merging this

### DIFF
--- a/RandBLAS/CMakeLists.txt
+++ b/RandBLAS/CMakeLists.txt
@@ -17,15 +17,15 @@ target_include_directories(RandBLAS INTERFACE
   $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
   $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/RandBLAS>)
 
-target_link_libraries(RandBLAS PUBLIC ${RandBLAS_libs})
+target_link_libraries(RandBLAS INTERFACE ${RandBLAS_libs})
 
 set(RandBLAS_cxx_opts -Wall -Wextra)
 if (SANITIZE_ADDRESS)
     list(APPEND RandBLAS_cxx_opts -fsanitize=address)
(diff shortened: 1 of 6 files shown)


rileyjmurray:
Ping @burlen: I resolved a merge conflict so that we could run CI. Not sure if I resolved in the appropriate way, but we're getting a CMake error with these changes.

RandBLAS/CMakeLists.txt
```diff
@@ -17,8 +17,15 @@ target_include_directories(RandBLAS INTERFACE
...
-target_link_libraries(RandBLAS INTERFACE ${RandBLAS_libs})
-target_compile_options(RandBLAS INTERFACE -Wall -Wextra)
+target_link_libraries(RandBLAS PUBLIC ${RandBLAS_libs})
+
+set(RandBLAS_cxx_opts -Wall -Wextra)
+if (SANITIZE_ADDRESS)
+    list(APPEND RandBLAS_cxx_opts -fsanitize=address)
+    target_link_options(RandBLAS PUBLIC -fsanitize=address)
+endif()
+
```
rileyjmurray, line 27: @burlen  CMake is angry that the ``INTERFACE`` specifications in ``target_link_libraries`` and ``target_compile_options`` has been replaced with ``PUBLIC``. It looks like changing these back to ``INTERFACE`` fixes things.

rileyjmurray:
I fixed a memory leak in ``test_sparse.cc`` identified by the address sanitizer. However, address sanitizer also identified nontrivial leaks in ``dense.hh``.

RandBLAS/CMakeLists.txt
burlen: yes, that was because of the change to header only (use INTERFACE) from a compiled library(use PUBLIC)

burlen:
@rileyjmurray leak fixed. macos failure is due to test tolerance not met

rileyjmurray (approved):

PR #34, opened 2023-02-08
test dense sketching for deterministic behavior when threading

burlen:
- adds a test to verify that the output is the same for 1 or n threads.
- better floating point compare
- better error messages from tests


rileyjmurray (approved):
Looks awesome!

rileyjmurray:
Waiting for the linux test to pass before merging.

PR #40, opened 2023-04-07
API design : make (n_rows, n_cols) the first two fields in structs that represent distributions

rileyjmurray:


rileyjmurray:
Finally! Caught all the bugs introduced by this change.

PR #44, opened 2023-04-28
DNM : submatrix generation

rileyjmurray:


rileyjmurray:
I just opened this to get a pretty git diff. We were getting compiler errors.

PR #48, opened 2023-05-12
GitHub Actions : test with openmp on macOS

rileyjmurray:


rileyjmurray:
@burlen I noticed that blaspp found OpenMP for C++ in the latest``openmp-macos`` build, while RandBLAS only found OpenMP for C. It might be worth looking at how blaspp tries to configure OpenMP in their build process:

https://github.com/icl-utk-edu/blaspp/blob/master/CMakeLists.txt#LL312C1-L324C8.

From what I can tell, the only difference is that they use ``find_package( OpenMP )`` instead of ``find_package( OpenMP COMPONENTS CXX)``.

burlen:
@rileyjmurray the default compiler doesn't have OpenMP support, the trick was to point to the right compiler. Should we integrate the changes in to core-macos workflow?

.github/workflows/core-macos.yml
```diff
@@ -8,32 +8,26 @@ jobs:
...
+
       - name: configure OS
         run: |
           # os level stuff
-          pwd
-          ls
           set -x
-          brew install googletest
-      
+          brew install googletest llvm
```
rileyjmurray, line 16: @burlen: do we need the change to ``llvm`` here? I was under the impression that we only needed llvm from homebrew if we wanted to compile with OpenMP. If the change isn't necessary, then I think it should be reverted here. I think there's value in keeping ``core-macos`` the simplest possible build configuration and relying on a separate build (``openmp-macos``) for the more complicated case with OpenMP.

PR #51, opened 2023-05-30
Kaiwen_branch: Submatrix generation

kaiwenhe7:
I added the function fill_rsubmat_omp that does the submatrix generation, modified fill_rmat to just call the new function. I also created a testing suite for the sub matrix generation that is exactly what I showed during our meeting on Friday.


rileyjmurray (changes requested):
This looks wonderful! I've asked you to make some changes related to documentation. Once you've made those changes I'll review in detail, but I expect we'll merge soon!

RandBLAS/dense.hh
```diff
@@ -204,88 +204,97 @@ DenseSkOp<T,RNG>::~DenseSkOp() {
...
+auto fill_rsubmat_omp(
     int64_t n_cols,
-    T* mat,
-    const RNGState<RNG> & seed,
-    MajorAxis ma = MajorAxis::Long
+    T* smat,
+    int64_t n_srows,
+    int64_t n_scols,
+    int64_t ptr,
+    const RNGState<RNG> & seed
```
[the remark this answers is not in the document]
kaiwenhe7: I have pushed the new changes related to the documentation. Please let me know what you think.

In the current implementation for fill_rsubmat_omp, the only information related to the original, larger random matrix is the number of columns. The implementation doesn't check if the sub matrix goes out of bounds of the original random matrix. Would you like fill_rsubmat_omp to have an additional argument for the number of rows of the original random matrix so that at the beginning we can do a check to make sure the submatrix is within bounds?

RandBLAS/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 
+set(CMAKE_CXX_STANDARD 20)
```
rileyjmurray, line 2: Please add a comment indicating why this is needed.

CMake comments start with ``#``.

RandBLAS/dense.hh
```diff
@@ -183,7 +183,98 @@ DenseSkOp<T,RNG>::~DenseSkOp() {
...
+ * @tparm OP an operator that transforms raw random values into matrix
+ *           elements. See r123ext::uneg11 and r123ext::boxmul.
+ *
+ * @param[in] n_cols the number of columns in the matrix.
+ * @param[in] smat a pointer to a contiguous region of memory with space for
+ *                n_rows \times n_cols elements of type T. This memory will be
+ *                filled with random values.
+ * @param[in] n_srows the number of rows in the submatrix.
+ * @param[in] n_scols the number of colomns in the submatrix.
+ * @param[in] ptr the starting locaiton within the random matrix, for which
```
rileyjmurray, line 202: This is assuming the random matrix is stored in row-major order?

rileyjmurray (commented):
I asked a couple questions. Once we converge on this this will be good to merge.

RandBLAS/dense.hh
```diff
@@ -183,7 +183,98 @@ DenseSkOp<T,RNG>::~DenseSkOp() {
...
+ * If RandBLAS is compiled with OpenMP threading support enabled, the operation is
+ * parallelized using OMP_NUM_THREADS. The sequence of values genrated is not
+ * dependent on the number of OpenMP threads.
+ *
+ * @tparam T the data type of the matrix
+ * @tparam RNG a random123 CBRNG type
+ * @tparm OP an operator that transforms raw random values into matrix
+ *           elements. See r123ext::uneg11 and r123ext::boxmul.
+ *
+ * @param[in] n_cols the number of columns in the matrix.
```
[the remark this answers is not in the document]
kaiwenhe7: Yes, that is correct

RandBLAS/dense.hh
kaiwenhe7: Yes. I think the function would also work in column major order if I add an additional argument that takes in the number of rows of the matrix. Should I make those changes? I would have to then do more testing to make sure its correct.

RandBLAS/CMakeLists.txt
kaiwenhe7: I think I added that because my default compiler had a c++ standard that threw errors with the use of auto in RandBLAS. I removed it and it seems to compile. I can remove that line.

RandBLAS/dense.hh
rileyjmurray: The original implementation more or less worked in column-major orientation. If anything, we'd want to switch to entirely column-major. But I suspect that the distinction between these cases isn't significant. Hold off on making any changes on this topic.

Context: when I recently updated RandBLAS to work in terms of "short-axis major" and "long-axis major" for dense sketching operators, I found that it suffices to have _either_ a row-major _or_  a column-major ``fill_rmat`` function, and that we can get the other format by playing with transposes.

kaiwenhe7: Okay that sounds good. Please let me know if there is anything else that needs to be addressed.

rileyjmurray: I'll go ahead and merge now. If I run into issues (and there's a decent chance that I will ...), then I'll ping you!

PR #55, opened 2023-06-05
Quality of life improvements : more overloading of sketching functions, revised web docs, and (partially) flattened namespaces

rileyjmurray:
The changes in this PR come in three broad categories

#### Namespaces
* RandBLAS::base has been flattened into the top-level RandBLAS namespace. 
* Functions for defining and filling sketching operators are now in the top-level RandBLAS namespace. 
* The RandBLAS::ramm namespace has been removed and flattened into the top-level RandBLAS namespace.

#### Documentation and general variable renaming in function signatures
 * The "abstract interface" part of the web docs has been renamed to "User Guide."
 * The "API" part of web docs has been removed. 
 * All instances of ``i_os`` and ``row_offset`` have been replaced by ``i_off``; similarly for ``j_os``.
 * All instances of ``transA`` were replaced by ``opA``; similarly for ``transS``.

#### Sparse sketching
* SparseDist::SparsityPattern has been replaced with RandBLAS::MajorAxis.
     * SparseSkOp.family has been replaced with SparseSkOp.major_axis.
     * This makes the SparseSkOp API closer to the DenseSkOp API.
     * SASOs are the same as what we now call short-axis major.
     * LASOs are the same as what we now call long-axis major.
 * The SparseSkOp struct now has overloaded constructors instead of one constructor with optional arguments.
 * A few utility functions in RandBLAS::sparse were moved into top-level RandBLAS.
 * The declaration order for properties of SparseDist has changed so that ``major_axis`` is last.
 *  The order of arguments for ``repeated_fisher_yates`` has changed.
 * Variables have been renamed in ``repeated_fisher_yates``  to speak in terms of the major-axis and minor-axis, rather than the "vec axis" and the "rep axis".
  * The ``lskges`` implementation replaced the ``pretend_layout_A`` variable name with the more informative ``layout_opA``.
 * The implementation of ``fill_sparse`` has been cleaned up.


rileyjmurray:
The build failure stems from BLAS++; its CMake isn't finding OpenMP on the basic macOS build (as expected) but then BLAS++ still tries to include ``omp.h`` at some point in its compilation process.

Issue #57, opened 2023-06-12
Generation of dense sketching operators doesn't return updated RNGState


rileyjmurray:
Ping @kaiwenhe7. It looks like your big PR a couple of weeks back introduced a bug. The current implementation (renamed to ``fill_dense_submat_impl`` a while back) can be found here https://github.com/BallisticLA/RandBLAS/blob/9debb57a34e07263f4c98356f740629cd1106121/RandBLAS/dense.hh#L203-L270

The old implementation can be found right around here:
https://github.com/BallisticLA/RandBLAS/commit/29ddfb7d182f990035352885a1721c6dec9bdfc4#diff-454044b3f0993309c60438575edec1d146c1cdfd03f6ca6002c59fb8729bec8cL208

The problem with the new implementation is that the RNGState it returns is based on ``{c, k}``, but ``c`` is never updated in the function. Compare this to the old implementation, which did update ``c``, e.g., here: 
https://github.com/BallisticLA/RandBLAS/commit/29ddfb7d182f990035352885a1721c6dec9bdfc4#diff-454044b3f0993309c60438575edec1d146c1cdfd03f6ca6002c59fb8729bec8cL267

rileyjmurray:
@TeachRaccooon is going to try working on this first.

kaiwenhe7:
Oh yes! In the parallel case, each thread creates its own copy of c. I forgot to update the actual counter at the end. Perhaps it should really simple to just advance c all the way to the final state at the end.

TeachRaccooon:
@kaiwenhe7 Guess my RandBLAS skills aren't strong enough, wasn't able to fix your code. 
If you know the solution, please also add a basic test for it so that we make sure this issue is fuly taken care of.

kaiwenhe7:
@TeachRaccooon The solution is more like a hack. Since the main body of the code only uses a private copy of c, the actual c is left untouched during the whole process of the sub matrix generation. Given the position of the last entry of the sub matrix n_scols*n_srows, the associated position of the counter can be computed. So before returning the state, I was thinking of just advancing the counter all the way to the final position.

TeachRaccooon:
@kaiwenhe7 do you mean just literally doing 
`    c.incr(n_srows * n_scols);
    
    return RNGState<RNG> {c, k};`?

kaiwenhe7:
@TeachRaccooon Yes. But I think it would be something like 
[...]

PR #58, opened 2023-06-12
Adding a fix and a simple test

TeachRaccooon:
Adding a fix (and a simple test) for issue #57 that has been affecting RandLAPACK.

### DIFF
--- a/test/test_dense/test_construction.cc
+++ b/test/test_dense/test_construction.cc
@@ -370,13 +370,66 @@ class TestStateUpdate : public ::testing::Test
 
         ASSERT_TRUE(!(A == B));
     }
+
+    template <typename T>
+    static void test_identity(
+        uint32_t key,
+        int64_t n_rows,
+        int64_t n_cols,
+        RandBLAS::DenseDistName dn
+    ) {
+        // Allocate workspace
+        int64_t size = n_rows * n_cols;
+        std::vector<T> A(size, 0.0);
+        std::vector<T> B(size, 0.0);
+
+        // Construct the sketching operator
+        RandBLAS::DenseDist D1 = {
+            .n_rows = n_rows,
+            .n_cols = n_cols / 2,
+            .family = dn
+        };
+
+        RandBLAS::DenseDist D2 = {
+            .n_rows = n_rows,
+            .n_cols = n_cols,
+            .family = dn
+        };
+
+        auto state = RandBLAS::RNGState(key);
+        // Concatenates two matrices generated from state and next_state
+        auto next_state = RandBLAS::fill_dense(D1, A.data(), state);
+        RandBLAS::fill_dense(D1, A.data() + (int64_t) ((float) size / 2.0), next_state);
+
+        char name [] = "A";
+        RandBLAS::util::print_colmaj(n_rows, n_cols, A.data(), name);
+
+        RandBLAS::fill_dense(D2, B.data(), state);
+
+        char name1 [] = "B";
+        RandBLAS::util::print_colmaj(n_rows, n_cols, B.data(), name1);
+
+        ASSERT_TRUE(A == B);
+    }
 };
 
 // For small matrix sizes, mean and stddev are not very close to desired vals.
-TEST_F(TestStateUpdate, Gaussian)
+TEST_F(TestStateUpdate, Gaussian_var_gen)
 {
     for (uint32_t key : {0, 1, 2}) {
         auto dn = RandBLAS::DenseDistName::Gaussian;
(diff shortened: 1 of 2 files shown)


TeachRaccooon:
@kaiwenhe7 Can you take a look at the failing test case?

kaiwenhe7:
Yes I can do that

kaiwenhe7:
I have a few things in my mind that may be causing the issue. I'm looking at the functions fill_dense and fill_dense_submat and for some reason I feel like there is some information being lost about the submatrix. I cloned the state_fix branch onto mine. I want to rewrite your test just using fill_dense_submat_impl, and I can get back to you.

kaiwenhe7:
@TeachRaccooon In your second call for generating the bottom half of the random matrix, I think you should use the function fill_dense_submat with i_off = 2 (since 50 mod 4 = 2 where 50 is the pointer for the start of the bottom half of the sub matrix and RNG::ctr_type_static_size = 4). When you call fill_dense, fill_dense calls full_dense_submat with i_off and j_off = 0, but the start of the new submatrix starts at index 2 of the array of random numbers associated with counter given by next_state.

test/test_dense/test_construction.cc
```diff
@@ -336,3 +336,100 @@ TEST_F(TestFillAxis, long_axis_2x4) {
...
+            .n_rows = n_rows,
+            .n_cols = n_cols / 2,
+            .family = dn
+        };
+
+        RandBLAS::DenseDist D2 = {
+            .n_rows = n_rows,
+            .n_cols = n_cols,
+            .family = dn
+        };
```
rileyjmurray, line 397: With the way that this test is currently being called, we have ``n_rows == n_cols``. That's a problematic edge-case when it comes to determining short-axis major versus long-axis major layout. We should first get this working with ``n_cols < n_rows``.

TeachRaccooon:
> @TeachRaccooon In your second call for generating the bottom half of the random matrix, I think you should use the function fill_dense_submat with i_off = 0 and j_off = 5. When you call fill_dense, fill_dense calls full_dense_submat with i_off and j_off = 0, but the start of the new submatrix starts at index 2 of the array of random numbers associated with counter given by next_state. I can test it myself and get back.

One quick note is that in this test, i am assuming column-major storage, so the second call will be filling the right half of the matrix.
 About the need to use `fill_dense_submat` on the subportion of a given matrix, shouldn't the user have an option to just call the regular `fill_dense` (for example, when column-major storage is used and numrows(sumbat) == numrows(mat))?
I am not seeing the reason why this should not work.
@rileyjmurray we spoke about something along these lines briefly yesterday, what are your thoughts? 

In any case, @kaiwenhe7, if you have a fix, please apply it.

kaiwenhe7:
@TeachRaccooon I have the fix on my branch. Should I open a pull request? 

There shouldn't be an issue if the user wants to call fill_dense to generate a random matrix one and done. However, in the concatenation, we should be calling fill_dense_submat at least for the second half because the second half is part of a larger random matrix. There needs to be additional information about the starting pointer of the sub matrix within the larger matrix that fill_dense does not take in.

TeachRaccooon:
> @TeachRaccooon I have the fix on my branch.
> 
> There shouldn't be an issue if the user wants to call fill_dense to generate a random matrix one and done. However, in the concatenation, we should be calling fill_dense_submat at least for the second half because the second half is part of a larger random matrix. There needs to be additional information about the starting pointer of the sub matrix within the larger matrix that fill_dense does not take in.

Why not just pass an offset pointer into `fill_dense`?
But anyway, we can discuss this later; for now i need this fix applied so that i can continue working on RandLAPACK.

You can either open a new PR or, presumably, just push your changes into this branch.

rileyjmurray:
@kaiwenhe7 

> However, in the concatenation, we should be calling fill_dense_submat at least for the second half because the second half is part of a larger random matrix. There needs to be additional information about the starting pointer of the sub-matrix within the larger matrix that fill_dense does not take in.

I don't fully agree with this. I want us to assume the user has an initial sketching operator ``S1``, and then they want a new sketching operator ``S2`` with the same aspect ratio (wide vs tall) and same major axis (short vs long) as ``S1``. If this happens, then the implicit matrix ``[S1; S2]`` (or ``[S1, S2]``, depending on dimensions) is the same as if we asked for a matrix ``S`` which is the same size as ``[S1; S2]`` (again, or ``[S1, S2]``) from the beginning.  Note that we're only asking that this be possible when ``(S1, S2)`` have the same aspect ratio _and_ the same major axis.

PR #59, opened 2023-06-13
Kaiwen branch

kaiwenhe7:
Fixed state bug of fill_dense_submat_impl and the test for test_identity.


rileyjmurray:
@kaiwenhe7, thanks for opening the PR, but you jumped the gun a bit by merging immediately. Please create a new PR so Max and I can review the tests and agree that things look good.

Issue #53, opened 2023-06-03
Adding rows or columns to dense sketching operators


rileyjmurray:
### Edit
Per Kaiwen's comment, this is actually quite easy. The main things users might still need are (1) populating $\texttt{S.next}\textunderscore\texttt{state}$ immediately when $\texttt{S}$ is constructed and (2) creating an API that doesn't require users to call ``fill_dense_submat``. Besides those things, we'd want to update the web documentation. Right now there's relevant docs on ["generating your N-th sketching operator, for N > 1"](https://randblas.readthedocs.io/en/latest/tutorial/sampling_skops.html#constructing-your-n-text-th-sketching-operator-for-n-1). It would be good to adapt the (very long!) content of my original comment below into proper web docs.

### Original content
RandBLAS should make it easy to _implicitly_ extend an initial sketching operator $S_1$ into an augmented operator $S = [S_1; S_2]$ or $S = [S_1, S_2]$.  There are four scenarios that a user can find themselves in where this can be done without generating $S$ from scratch. I've provided them for completeness below.

_Remark 1._ In all four scenarios, the idea is that we should be able to use the $\texttt{next}\textunderscore\texttt{state}$ of $S_1$ as the $\texttt{seed}\textunderscore\texttt{state}$ of $S_2$. Our current implementation of $\texttt{DenseSkOp}$ doesn't set $\texttt{S.next}\textunderscore\texttt{state}$ until the user calls ``fill_dense`` on $\texttt{S}$. So at the very least, we need to change our code so that $\texttt{S.next}\textunderscore\texttt{state}$ is set the moment that $\texttt{S}$ is constructed.

_Remark 2._ The four scenarios below show two reasons why you'd want to extend a sketching operator; you might be trying to improve statistical quality by increasing sketch size (Scenarios 1 and 4), or you might be incorporating new data into a sketch of fixed size (Scenarios 2 and 3). The unifying perspective on Scenarios 1 and 4 is that they both add long-axis vectors to the sketching operator. The unifying perspective on Scenarios 2 and 3 is that they both add short-axis vectors to the sketching operator. $\texttt{DenseDist}$ objects have a $\texttt{major}\textunderscore\texttt{axis}$ member, which states whether operators sampled from that distribution are short-axis or long-axis major. So when a user specifies the major axis for a sketching operator, they're basically saying whether they might care about improving the statistical quality of a sketch or updating a sketch to incorporate more data.

[...]

PR #62, opened 2023-06-13
Bugfixes for generating dense sketching operators

kaiwenhe7:
This PR resolves the bug from #57 in a way that supports the functionality mentioned in #53.

### DIFF
--- a/test/test_dense/test_construction.cc
+++ b/test/test_dense/test_construction.cc
@@ -525,7 +525,7 @@ class TestStateUpdate : public ::testing::Test
     }
     
     template<typename T>
-    static void test_identity_v2(
+    static void test_identity_trunc(
         uint32_t key,
         int64_t n_rows,
         int64_t n_cols
@@ -541,12 +541,11 @@ class TestStateUpdate : public ::testing::Test
         int64_t ptr = n_srows1 * n_cols;
 
         auto state = RandBLAS::RNGState(key);
-        auto stater = RandBLAS::RNGState(key);
-        auto state1 = RandBLAS::RNGState(key);
-        auto next_state = RandBLAS::fill_dense_submat_impl<T, r123::Philox4x32,r123ext::boxmul>(n_cols, A.data(), n_srows1, n_cols, 0, state);
-        RandBLAS::fill_dense_submat_impl<T, r123::Philox4x32,r123ext::boxmul>(n_cols, A.data() + ptr, n_srows2, n_cols, n_srows1*n_cols  % 4, next_state);
 
-        RandBLAS::fill_dense_submat_impl<T, r123::Philox4x32,r123ext::boxmul>(n_cols, B.data(), n_rows, n_cols, 0, state);
+        auto next_state = RandBLAS::fill_dense_submat_trunc<T, r123::Philox4x32,r123ext::boxmul>(n_cols, A.data(), n_srows1, n_cols, 0, state);
+        RandBLAS::fill_dense_submat_trunc<T, r123::Philox4x32,r123ext::boxmul>(n_cols, A.data() + ptr, n_srows2, n_cols, 0, next_state);
+
+        RandBLAS::fill_dense_rmat_trunc<T, r123::Philox4x32,r123ext::boxmul>(B.data(), n_rows, n_cols, state);
 
         for (int i = 0; i < size; i++) {
             total += abs(A[i] - B[i]);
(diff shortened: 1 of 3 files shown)


rileyjmurray:
Thanks for re-opening the PR! My first comment concerns the tests. Please remove unnecessary print statements and please enable tests that are currently commented-out.

rileyjmurray:
@kaiwenhe7 you recently left a comment on PR #57 mentioning that you've tested matrices whose dimensions are prime numbers. Please push those test changes to this branch so that we can inspect them.

rileyjmurray:
Thanks for that change, @kaiwenhe7. Please also address my comments from earlier:

https://github.com/BallisticLA/RandBLAS/pull/62#issuecomment-1589947904

kaiwenhe7:
@rileyjmurray @TeachRaccooon I am currently working to try and resolve the uncommented tests.

TeachRaccooon:
> @rileyjmurray @TeachRaccooon I am currently working to try and resolve the uncommented tests. Is it okay if I rewrite the concatenation tests just using fill_dense_submat_impl? I don't fully grasp all of the moving parts of fill_dense_submat. I am currently stuck on the debugging, and it may be good for me to simplify it just using fill_dense_submat_impl.

Sure, whatever works!

TeachRaccooon:
@kaiwenhe7 any updates?

kaiwenhe7:
@TeachRaccooon I'm struggling with some things regarding rounding when we're portioning the matrix. I have it working for when the dimensions are even. When we we're splitting the matrix using n_cols / 2, I may be messing up some rounding stuff.

kaiwenhe7:
@TeachRaccooon There is something interesting from doing more testing. I copied the test_identity function to a separate c++ file and when I ran the exact same test, it seems to not have any errors. Here are some screen shots. If any entry of A and B are not equal the values would be printed out, but when running the file, nothing is printed, which means that each entry should be bitwise equal. On the other hand, the exact same function fails the test in the testing suite.

Screen Shot 2023-06-14 at 4 45 05 PM
Screen Shot 2023-06-14 at 4 45 22 PM

kaiwenhe7:
@rileyjmurray I have the truncated version of the submatrix generation implemented along with the exact same tests from the previous implementation. One last choice that needs to be made is the state of the counter than needs to be returned at the end. If we decide to return the state corresponding to the last element of the submatrix, if we want to add rows to the matrix, the user would need to advance to the next counter and then pass it in. Alternatively, we could ourselves by advancing the counter an additional time before returning the state.

rileyjmurray:
Excellent! I'd like for us to advance the counter one additional time automatically, rather than leave that to the user.

rileyjmurray:
(Sorry for accidentally closing the PR for a second there.)

kaiwenhe7:
@rileyjmurray I have the correct state returned and I modified test_identity for the sub matrix concatenation to use the truncated implementation. In the meantime, I will work on the method of returning the correct state that avoids the risk of overflow from the ptr.

kaiwenhe7:
@rileyjmurray Should I delete the old implementations of the submatrix generation?

rileyjmurray:
@kaiwenhe7 if the new implementation is tested at least as thoroughly as the old implementation, then yes.

kaiwenhe7:
@rileyjmurray I made all of the changes you requested in our last meeting. In regards to the counter stuff, maybe that can be done in a separate pull request? Let me know what you think of the current changes.

test/test_dense/test_construction.cc
```diff
@@ -336,3 +337,139 @@ TEST_F(TestFillAxis, long_axis_2x4) {
...
+        RandBLAS_Testing::Util::fill_dense_rmat_trunc<T, r123::Philox4x32,r123ext::boxmul>(B.data(), n_rows, n_cols, state);
+
+        for (int i = 0; i < size; i++) {
+            total += abs(A[i] - B[i]);
+        }
+
+        ASSERT_TRUE(total == 0);
+
+    }*/
+};
```
rileyjmurray, line 454: @kaiwenhe7 is this test commented out deliberately?

rileyjmurray:
@kaiwenhe7 I left a comment about a test that's been commented out. Besides that, I agree that we can leave the delicate counter incrementing to the next PR.

test/test_dense/test_construction.cc
kaiwenhe7: @rileyjmurray Yes. I commented it out because the original test_identity works now using fill_dense.

rileyjmurray: Deleting it would be the right move here

kaiwenhe7: Cool done!

PR #63, opened 2023-07-04
Dense sketching operators: avoid int overflow when returning RNGState

kaiwenhe7:


rileyjmurray:
@kaiwenhe7 the tests are failing on macOs when OpenMP isn't installed: https://github.com/BallisticLA/RandBLAS/actions/runs/5452751148/jobs/9920682956?pr=63.

kaiwenhe7:
@rileyjmurray Is there a way of disabling openmp so that I can do testing in the case that a machine does not have openmp installed? I don't want to try and mess with the cake files too much.

rileyjmurray:
@kaiwenhe7 use
```
#if defined(RandBLAS_HAS_OpenMP)
  ... code ...
#endif
```

rileyjmurray:
Okay, the basic build passes with this change, but the address sanitizer build fails. Compiler warnings for the address sanitizer build show that some values (``ctr_arr``, ``thrd``) aren't being initialized properly; see [here](https://github.com/BallisticLA/RandBLAS/actions/runs/5465964587/jobs/9950195418?pr=63#step:7:44).

kaiwenhe7:
@rileyjmurray I think it should be fixed now. The preexisting tests pass, which means that its returning the correct counter, but I am writing additional tests to make sure.

rileyjmurray:
Ah shoot, in that case I may have jumped the gun. Please make a new PR with any tests you'd like.

PR #74, opened 2023-09-01
Implementation of sketch_vector

kaiwenhe7:
Created sketch_vector function in skge.hh and added unit test test_sketch_vec

### DIFF
--- a/RandBLAS/skge.hh
+++ b/RandBLAS/skge.hh
@@ -596,6 +596,23 @@ void sketch_general(
     return sketch_general(layout, opA, opS, m, d, n, alpha, A, lda, S, 0, 0, beta, B, ldb);
 };
 
+template <typename T, typename SKOP>
+void sketch_vector(
+    blas::Op opS,
+    int64_t d, // length of y
+    int64_t m, // length of x
+    T alpha,
+    SKOP &S,   // S has dim d by m
+    int64_t i_off,
+    int64_t j_off,
+    const T *x,
+    int64_t incx,
+    T beta,
+    T *y,
+    int64_t incy
+) {
+    return sketch_general(blas::Layout::RowMajor, opS, blas::Op::NoTrans, d, 1, m, alpha, S, i_off, j_off, x, incx, beta, y, incy);
+}
 
 }  // end namespace RandBLAS
 #endif
--- a/RandBLAS/skge.hh
+++ b/RandBLAS/skge.hh
@@ -603,7 +603,7 @@ void sketch_general(
 /// Perform a GEMV-like operation
 /// @verbatim embed:rst:leading-slashes
 /// .. math::
-///     \mat(y) = \alpha \cdot \underbrace{\op(S)}_{d \times m} \cdot \underbrace{\mat(x)}_{m \times 1} + \beta \cdot \underbrace{\mat(y)}_{d \times 1},    \tag{$\star$}
+///     \mat(y) = \alpha \cdot \underbrace{\op(\submat(S))}_{d \times m} \cdot \underbrace{\mat(x)}_{m \times 1} + \beta \cdot \underbrace{\mat(y)}_{d \times 1},    \tag{$\star$}
 /// @endverbatim
 /// where \math{\alpha} and \math{\beta} are real scalars and \math{S} is a sketching operator.
 /// 
@@ -611,7 +611,10 @@ void sketch_general(
 /// What are :math:`\mat(x)` and :math:`\mat(y)`?
 ///     Their shapes are defined as tall vectors of dimension :math:`(\mat(x), m \times 1)`, :math:`(\mat(y), d \times 1)`.
 ///     Their precise contents are determined by :math:`(x, incx)`, :math:`(y, incy)`,
-///     and "layout = RowMajor", following the same convention as BLAS.
+///     and "layout = RowMajor", following the same convention as BLAS. We note that this representation is unusual, but it allows us to interpret incx and incy like lda in GEMM.
+///     This function treats (\mat{x},incx) and (\mat{x}, incy) the exact same way as Netlib BLAS, which uses column-major layout. In interfaces like CBLAS, which allow either row-major or column-major
+///     layout, the layout argument in GEMV only affects the matrix. sketch_vector does not have a Layout argument because the matrix is a sketching operator,
+///     and RandBLAS does not deal with the "layout" of a sketching operator in the conventional way.
 /// @endverbatim
 ///
 /// @param[in] opS
@@ -624,7 +627,7 @@ void sketch_general(
 ///
 /// @param[in] m
 ///     A nonnegative integer.
-///     - The number of elements of \math{\mat(B)}
+///     - The number of elements of \math{\mat(y)}
 ///
 /// @param[in] alpha
 ///     A real scalar.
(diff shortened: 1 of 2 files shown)


kaiwenhe7:
@rileyjmurray Hi Riley. I finished the testing that you requested on Friday. I also added the documentation for sketch_vector. I'm unsure how to see if it looks good. Let me know what you think. Thanks!

RandBLAS/skge.hh
```diff
@@ -596,6 +596,84 @@ void sketch_general(
...
+/// .. math::
+///     \mat(y) = \alpha \cdot \underbrace{\op(S)}_{d \times m} \cdot \underbrace{\mat(x)}_{m \times 1} + \beta \cdot \underbrace{\mat(y)}_{d \times 1},    \tag{$\star$}
+/// @endverbatim
+/// where \math{\alpha} and \math{\beta} are real scalars and \math{S} is a sketching operator.
+/// 
+/// @verbatim embed:rst:leading-slashes
+/// What are :math:`\mat(x)` and :math:`\mat(y)`?
+///     Their shapes are defined as tall vectors of dimension :math:`(\mat(x), m \times 1)`, :math:`(\mat(y), d \times 1)`.
+///     Their precise contents are determined by :math:`(x, incx)`, :math:`(y, incy)`,
+///     and "layout = RowMajor", following the same convention as BLAS.
```
rileyjmurray, line 614: This is a little confusing. I get that since you've declared them as column vectors, the increment is effectively a stride in a row-major representation, but this is not the usual way of describing things.

I suggest that you explicitly acknowledge this presentation as unusual. Say that we choose this representation so that ``incx`` and ``incy`` can be interpreted like ``lda`` in GEMM. Emphasize that this function treats (x, incx), (y, incy) in the exact same way as _Netlib_ BLAS (which, of course, uses column-major layout). For interfaces like CBLAS which allow either row-major or column-major layout, the layout argument in GEMV only affects the matrix (see [here](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2023-0/cblas-gemv.html) for an example). Explain that we have no ``Layout`` argument in ``sketch_vector`` because our matrix is a sketching operator, and RandBLAS never speaks of the "layout" of a sketching operator.

^ That's pretty verbose, but you get the idea.

RandBLAS/skge.hh
```diff
@@ -596,6 +596,84 @@ void sketch_general(
...
+///      - If \math{\opS} = NoTrans, then \math{ \op(S) = S}.
+///      - If \math{\opS} = Trans, then \math{\op(S) = S^T }.
+///
+/// @param[in] d
+///     A nonnegative integer.
+///     - The number of elements of \math{\mat(x)}.
+///
+/// @param[in] m
+///     A nonnegative integer.
+///     - The number of elements of \math{\mat(B)}
```
rileyjmurray, line 627: Typo; there's no ``B``.

RandBLAS/skge.hh
```diff
@@ -596,6 +596,84 @@ void sketch_general(
...
+///     A real scalar.
+///     - If zero, then \math{x} is not accessed.
+///
+/// @param[in] x
+///     Pointer to a 1D array of real scalars.
+///     - Defines \math{\mat(x)}.
+///
+/// @param[in] incx
+///     A nonnegative integer.
+///     * Stride between elements of x. incx must not be zero.
```
rileyjmurray, line 639: Clarify. Can it be negative? Don't actually think ``sketch_general`` supports negative values for LDA ... but GEMV in BLAS does allow negative incx...

RandBLAS/skge.hh
```diff
@@ -596,6 +596,84 @@ void sketch_general(
...
+///     
+/// @param[in] S
+///    A DenseSkOp or SparseSkOp object.
+///    - Defines \math{S}.
+///
+/// @param[in] beta
+///     A real scalar.
+///     - If zero, then \math{x} need not be set on input.
+///
+/// @param[in, out] B
```
rileyjmurray, line 649: Should be ``y``, not ``B``.

test/test_dense/test_sketch_gefl3.cc
```diff
@@ -210,8 +210,195 @@ class TestLSKGE3 : public ::testing::Test
...
+        T y_ref[incy*d];
+        
+        // Initialize elemnts of x to 1 with stride of incx between elements
+        for (int i = 0; i < incx*m; i++){
+            x[i] = 1.0;
+        }
+
+        // Generate wide sketching operator
+        RandBLAS::DenseDist D(d, m);
+        RandBLAS::DenseSkOp<double> S_wide(D, seed, nullptr);
```
rileyjmurray, line 233: If you template on ``T`` then you should use ``RandBLAS::DenseSkOp<T>``, not ``RandBLAS::DenseSkOp<double>``.

RandBLAS/skge.hh
```diff
@@ -596,6 +596,84 @@ void sketch_general(
     return sketch_general(layout, opA, opS, m, d, n, alpha, A, lda, S, 0, 0, beta, B, ldb);
 };
 
+// =============================================================================
+/// \fn sketch_vector(blas::Op opS, int64_t d, int64_t m, T alpha, SKOP &S,
+///    int64_t i_off, int64_t j_off, const T *x, incx, T beta, T *y, incy
+/// )
```
rileyjmurray, line 602: As a convenience function, I think we should offer ``sketch_vector`` that omits ``i_off`` and ``j_off`` and just sets them to zero. You don't need to test this function separately. The documentation can be copied from what you end up writing here (make sure we agree on it first!), and then you can trim down the copied text.

RandBLAS/skge.hh
```diff
@@ -596,6 +596,84 @@ void sketch_general(
...
+/// Perform a GEMV-like operation
+/// @verbatim embed:rst:leading-slashes
+/// .. math::
+///     \mat(y) = \alpha \cdot \underbrace{\op(S)}_{d \times m} \cdot \underbrace{\mat(x)}_{m \times 1} + \beta \cdot \underbrace{\mat(y)}_{d \times 1},    \tag{$\star$}
+/// @endverbatim
+/// where \math{\alpha} and \math{\beta} are real scalars and \math{S} is a sketching operator.
```
rileyjmurray, line 608: We should be using ``\op(\submat(S))``, not ``\op(S)``. The fact that we allow ``i_off, j_off`` to be nonzero is the difference between plain ``\op(S)`` versus ``\op(\submat(S))``.

RandBLAS/skge.hh
```diff
@@ -596,6 +596,84 @@ void sketch_general(
...
+///     Pointer to a 1D array of real scalars.
+///     - Defines \math{\mat(x)}.
+///
+/// @param[in] incx
+///     A nonnegative integer.
+///     * Stride between elements of x. incx must not be zero.
+///     
+/// @param[in] S
+///    A DenseSkOp or SparseSkOp object.
+///    - Defines \math{S}.
```
rileyjmurray, line 643: You're missing documentation for ``i_off`` and ``j_off``.

test/test_dense/test_sketch_gefl3.cc
rileyjmurray, line 1: These tests are beautiful! Well done, @kaiwenhe7.

rileyjmurray (changes requested):
Thanks for your patience, @kaiwenhe7. I've left a large number of comments about the documentation for ``sketch_vector``. I'd also like you to make a wrapper for ``sketch_vector`` which automatically sets ``i_off = j_off = 0`` (you can't do this with default arguments since there are arguments without defaults that follow).

PR #75, opened 2023-11-10
Added new folder for examples of sketching

kaiwenhe7:
Inside the folder is a cmakelists.txt that requires RandBLAS, Blaspp, and Lapackpp. I have two implementations of TLS. One uses dense sketching and the other uses sparse sketching.

### DIFF
--- a/examples/TLS_DenseSkOp.cpp
+++ b/examples/TLS_DenseSkOp.cpp
@@ -1,0 +1,166 @@
+#include <blas.hh>
+#include <RandBLAS.hh>
+#include <lapack.hh>
+
+#include <omp.h>
+#include <stdio.h>
+#include <unistd.h>
+#include <iostream>
+#include <math.h>
+#include <time.h>
+#include <stdlib.h>
+#include <chrono>
+
+using std::chrono::high_resolution_clock;
+using std::chrono::duration_cast;
+using std::chrono::duration;
+using std::chrono::milliseconds;
+
+
+//TODO: Read in matrix dimensions 
+
+void init_noisy_data(int64_t m, int64_t n, int64_t d, double* AB){
+    double target_x[n*d];
+    double eps[m*d];
+    for (int i = 0; i < n; i++) {  
+        target_x[i] = 1;   // Target X is the vector of 1's
+    }
+
+    RandBLAS::DenseDist Dist_A(m,n); 
+    RandBLAS::DenseDist Dist_eps(m,d); 
+    auto state = RandBLAS::RNGState(0);
+    auto state1 = RandBLAS::RNGState(1);
+
+    RandBLAS::fill_dense<double>(Dist_A, AB, state);  //Fill A to be a random gaussian
+    RandBLAS::fill_dense<double>(Dist_eps, eps, state1);  //Fill A to be a random gaussian
+
+    blas::gemm(blas::Layout::ColMajor, blas::Op::NoTrans, blas::Op::NoTrans, m, d, n, 1, AB, m, target_x, n, 0, &AB[m*n], m);
+
+    for (int i = 0; i < m*d; i++){
+        AB[m*n + i] += eps[i];   // Add Gaussian Noise to right hand side
+    }
+}
+
+int main(int argc, char* argv[]){
+    // Goal: Solve total least squares problem ||AX - B|| 
+
+    // Initialize dimensions
+    int64_t m;           // Number of rows of A, B
+    int64_t n;           // Number of columns of A
+    if (argc == 3) {
+        m = atoi(argv[1]);
+        n = atoi(argv[2]);
+        if (n > m) {
+            std::cout << "Make sure number of rows are greater than number of cols" << '\n';
+            exit(0);
+        }
+    } else {
+        m = 10000;
+        n = 500;
+    }
+    int64_t d = 1;       // Number of columns of B
+    int64_t sk_dim = 2*(n+d);
+
+    // Initialize workspace
+    double *AB = new double[m*(n + d)]; // Store [A B] in column major format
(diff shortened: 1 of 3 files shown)


examples/TLS_DenseSkOp.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+    blas::gemm(blas::Layout::ColMajor, blas::Op::NoTrans, blas::Op::NoTrans, m, d, n, 1, AB, m, target_x, n, 0, &AB[m*n], m);
+
+    for (int i = 0; i < m*d; i++){
+        AB[m*n + i] += eps[i];   // Add Gaussian Noise to right hand side
+    }
+}
+
+int main(int argc, char* argv[]){
+    // Goal: Solve total least squares problem ||AX - B||
```
rileyjmurray, line 43: @kaiwenhe7 This is a little misleading. Please write a brief description of the total least squares problem. Please also explain how to call this executable.

examples/TLS_DenseSkOp.cpp
```diff
@@ -0,0 +1,164 @@
...
+        m = atoi(argv[1]);
+        n = atoi(argv[2]);
+        if (n > m) {
+            std::cout << "Make sure number of rows are greater than number of cols" << '\n';
+            exit(0);
+        }
+    } else {
+        m = 10000;
+        n = 500;
+    }
```
rileyjmurray, line 58: How about: if argc is 1, then m=10000 and n=500. If argc is 3, then you read input as above. In other situations you throw an error message and exit with code 1.

examples/TLS_DenseSkOp.cpp
```diff
@@ -0,0 +1,164 @@
...
+            1,                         // Scalar alpha - if alpha is zero AB is not accessed
+            S,                         // A DenseSkOp or SparseSkOp sketching operator
+            0,                         // Row offset of S
+            0,                         // Column offset of S
+            AB,                        // Matrix to be sketched
+            m,                         // Leading dimension of AB
+            0,                         // Scalar beta - if beta is zero SAB is not accessed
+            SAB,                       // Sketched matrix SAB
+            sk_dim                     // Leading dimension of SAB
+    );
```
rileyjmurray, line 117: You can remove the row and column offsets of S. We have an overload of ``sketch_general`` that automatically sets them to 0 if not provided.

examples/TLS_DenseSkOp.cpp
```diff
@@ -0,0 +1,164 @@
...
+    // Perform SVD operation on SAB
+    auto time_TLS1 = high_resolution_clock::now();
+    lapack::gesdd(lapack::Job::AllVec, sk_dim, (n+d), SAB, sk_dim, svals, U, sk_dim, VT, n+d);
+            
+    for (int i = 0; i < n; i++) {
+        X[i] = VT[(n+d-1) + i*(n+d)]; // Take the right n by 1 block of V
+    }
+
+    // Scale X by the inverse of the 1 by 1 bottom right block of V
+    blas::scal(n, -1/VT[(n+d)*(n+d)-1], X, 1);
```
rileyjmurray, line 129: This implementation is only valid when ``d = 1``. With that in mind it's probably best to eliminate the symbol ``d`` entirely and just replace it directly with 1 when needed.

examples/TLS_DenseSkOp.cpp
```diff
@@ -0,0 +1,164 @@
...
+            
+    for (int i = 0; i < n; i++) {
+        X[i] = VT[(n+d-1) + i*(n+d)]; // Take the right n by 1 block of V
+    }
+
+    // Scale X by the inverse of the 1 by 1 bottom right block of V
+    blas::scal(n, -1/VT[(n+d)*(n+d)-1], X, 1); 
+    auto time_TLS2 = high_resolution_clock::now();
+
+    //Check TLS solution. Expected to be a vector of 1's
```
rileyjmurray, line 132: Note: you expect it to be _close to_ a vector of all ones.

examples/TLS_DenseSkOp.cpp
```diff
@@ -0,0 +1,164 @@
...
+        }
+    }
+
+    std::cout << "Matrix dimensions:                              " << m << " by " << n+d << '\n'; 
+    std::cout << "Sketch dimension:                               " << sk_dim << '\n'; 
+    std::cout << "Time to create dense sketch:                    " << (double) duration_cast<milliseconds>(time_constructsketch2 - time_constructsketch1).count()/1000 << " seconds" << '\n';
+    std::cout << "Time to sketch AB:                              " << (double) duration_cast<milliseconds>(time_sketch2 - time_sketch1).count()/1000 << " seconds" <<'\n';
+    std::cout << "Time to perform TLS on sketched matrix:         " << (double) duration_cast<milliseconds>(time_TLS2 - time_TLS1).count()/1000 << " seconds" << '\n';
+    std::cout << "Inf-norm of TLS residal vector:                 " << res_infnorm << '\n';
+    std::cout << "Two-norm of TLS residual vector:                " << sqrt(res_twonorm) << '\n';
```
rileyjmurray, line 150: The vector ``X - expected_X`` isn't what's usually meant by "residual vector" in the context of  least squares (total, or otherwise).

Suggestion: report "inf-norm distance from TLS solution to vector of all ones" and "two-norm distance from TLS solution to vector of all ones".

rileyjmurray (changes requested):
Thanks for the PR, @kaiwenhe7!

I've left requests for changes in ``TLS_DenseSkOp.cpp``. Please address when you get a chance. Naturally, any changes in this file should also apply to ``TLS_SparseSkOp.cpp``.

rileyjmurray (approved):
This looks perfect! Thanks so much, @kaiwenhe7!

PR #101, opened 2024-07-03
Infrastructure for Kolmogorov-Smirnov tests, (tested!) methods for sampling with replacement, and changes to include guards

rileyjmurray:
As usual, the scope of this PR exceeded my original plans.

### Incidental, but notable changes

I removed the include-guard pattern
```c++
#ifndef <file identifier>_hh
#define <file identifier>_hh
// ... code ...
#endif
```
and replaced it with ``#pragma once`` everywhere.

### Main changes

I added three functions to ``RandBLAS/util.hh``:
* sample_indices_iid: sample with replacement from an index set according to a (nonuniform) probability distribution, as specified by its cumulative distribution function.
* weights_to_cdf: converts a buffer of nonnegative numbers into a cumulative distribution function.
* sample_indices_iid_uniform: a more efficient version of sample_indices_iid, specialized to sampling from the uniform distribution.

I added a file
> test/test_basic_rng/rng_common.hh

This file is where we'll put code that's used to construct our statistical tests. Conceptually distinct parts of the file:
 * a hard-coded statistical table for running two-sided Kolmogorov-Smirnov tests, plus functions for performing lookups in this table.
 * a section for purely combinatorial helper functions. Right now there's only one such function:``log_binomial_coefficient``.
 * a section for computing some quantities of interest for the hypergeometric distribution. The function for constructing the PMF is important since we can use it in a Kolmogorov-Smirnov test for correctness of ``repeated_fisher_yates`` (our function for sampling uniformly from an index set without replacement). Note: I haven't implemented this test yet!
 
I also added
> test/test_basic_rng/test_sample_indices.cc

Right now it only contains tests for sampling _with replacement_. It should also contain tests for sampling _without replacement_, where an empirical CDF for the hypergeometric distribution can be constructed from ``repeated_fisher_yates`` and the true CDF can be computed from ``test_basic_rng/rng_common.hh``.

### DIFF
--- a/RandBLAS/base.hh
+++ b/RandBLAS/base.hh
@@ -27,8 +27,7 @@
 // POSSIBILITY OF SUCH DAMAGE.
 //
 
-#ifndef randblas_base_hh
-#define randblas_base_hh
+#pragma once
 
 /// @file
 
@@ -212,5 +211,3 @@ std::ostream &operator<<(
 }
 
 } // end namespace RandBLAS::base
-
-#endif
--- a/RandBLAS/base.hh
+++ b/RandBLAS/base.hh
@@ -53,9 +53,18 @@
 /// code common across the project
 namespace RandBLAS {
 
+/**
+ * Stores stride information for a matrix represented as a buffer.
+ * The intended semantics for a buffer "A" and the conceptualized
+ * matrix "mat(A)" are 
+ * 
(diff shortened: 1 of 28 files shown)


rileyjmurray:
TODO: make weights_to_cdf robust to numerical issues, like the possibility of dividing by zero or having (ever so slightly) negative elements.

PR #106, opened 2024-08-04
KS-Tests for Fisher-Yates

RylieWeaver:
This PR is being created as a draft, serving as tracker for current changes until ready for review.


test/test_basic_rng/rng_common.hh
```diff
@@ -195,3 +210,53 @@ double hypergeometric_variance(int64_t N, int64_t K, int64_t D) {
...
+    for (size_t i = 0; i < cdf1.size(); ++i)
+    {
+        double diff = std::abs(cdf1[i] - cdf2[i]);
+        if (diff > max_diff)
+        {
+            max_diff = diff;
+        }
+    }
+    return max_diff;
+}
```
rileyjmurray, line 236: It's preferable for us to know the first index where the test would fail. One way forward is to change this function to accept a tolerance and return a ``std::tuple<int, double>`` of the form ``{i, first_violation}`` where ``i`` is the first index where ``first_violation = std::abs(cdf1[i] - cdf2[i])`` exceeds the tolerance.

rileyjmurray:
Hi, @RylieWeaver. Thanks so much for your contribution, and welcome to the project! 

I've left one small comment so far. I'd like to try and get this merged in quickly if possible. I can have fast turn-around on questions you might ask here on Sunday (August 4). After that I'll only be available in the evenings.

test/test_basic_rng/rng_common.hh
RylieWeaver: Hmm, ok. This is not so much what I expected. Is the intuition here that sharing '0' elements should be very close to real randomness, whereas sharing 'k' elements will almost surely not represent real randomness? Meaning, the goal of the function is to tell when this sort of switch from random to not random happens? (according to some threshold)

RylieWeaver:
> Hi, @RylieWeaver. Thanks so much for your contribution, and welcome to the project!
> 
> I've left one small comment so far. I'd like to try and get this merged in quickly if possible. I can have fast turn-around on questions you might ask here on Sunday (August 4). After that I'll only be available in the evenings.
> 
> _Edit._ You'll note that I have another PR open that includes tests. I'm happy to merge in your PR first. However, before we do that, you'll actually need to fix some bugs that I unearthed in my other PR. Please make the changes indicated here:
> 
> * https://github.com/BallisticLA/RandBLAS/pull/104/files#diff-dd85520be032a7f810e41e7e0a85c2baefab04f1fa508f93fb7da10b875f87abL114 (changes an inequality from `<=` to `>=`)
> * https://github.com/BallisticLA/RandBLAS/pull/104/files#diff-a034d04bfcd2668b3598b46e72b6a0985afc0b2eab283da40df6926707a079c6R74-R79 (fixes check of the KS test statistic. You actually wrote a new function for this, so you don't need the fix verbatim).
> * https://github.com/BallisticLA/RandBLAS/pull/104/files#diff-a034d04bfcd2668b3598b46e72b6a0985afc0b2eab283da40df6926707a079c6R98 (fixes an incorrectly-specified type).

Thanks! I'm excited to work on the project. I'm not yet an expert on C++ but looking forward to improving as I contribute.

Gotcha. I'll plan to directly copy-paste those changes, but let me know if you think making a sub-branch of that branch would be better.

test/test_basic_rng/rng_common.hh
rileyjmurray: My initial comment probably wasn't clear. I'm just suggesting that you modify the function as follows
```c++
std::pair<int,double> check_ks_below_critval(const std::vector<double> &cdf1, const std::vector<double> &cdf2, double critical_value)
{
    assert(cdf1.size() == cdf2.size()); // Vectors must be of same size to perform test

    for (size_t i = 0; i < cdf1.size(); ++i) {
        double diff = std::abs(cdf1[i] - cdf2[i]);
        if (diff > critical_value) {
            return {i, diff}; // the test failed.
        }
    }
    return {-1, 0.0};  // interpret a negative return value as the test passing.
}
```
Presumably you were computing ``ks_stat`` and then checking if it exceeded a given critical value. I'm suggesting that you do both steps in one function. This is useful because if the test fails then we immediately can get information that's useful for debugging (the index i where the empirical and true CDFs differ by more than the critical value).

Does that make sense?

RylieWeaver: I see, this does make sense. Yep, you have the right assessment of what I was doing before. I was approaching it from more of a "one-time test" where we could see a grid of significance values, but see now that's it's more of a debugging process.

RylieWeaver:
@rileyjmurray I'm getting a bit of a weird output that the hypergeometric_value returns nan. For example, with a simple d=9, k=7, observed_k=5, I get the following:

± |main U:1 ✗| → ./fisher_test 
Empirical CDF: 0 0 0 0 0 0.4375 0.96875 1 
Hypergeometric PMF for d=9, k=7, observed_k=5: nan

I am calling it like this:
`double pmf_test = RandBLAS_StatTests::hypergeometric_pmf(d, d, k, 5);
    std::cout << "Hypergeometric PMF for d=" << d << ", k=" << k << ", observed_k=5: " << pmf_test << std::endl;`

test/test_basic_rng/test_sample_indices.cc
```diff
@@ -88,13 +89,13 @@ class TestSampleIndices : public ::testing::Test
 
         RNGState state(seed);
         std::vector<int64_t> samples(num_samples, -1);
-        RandBLAS::util::sample_indices_iid_uniform(N, samples.data(), num_samples, state);
+        RandBLAS::util::sample_indices_iid_uniform(N, num_samples, samples.data(), state);
```
rileyjmurray, line 92: This change should be accompanied by a change to the definition of ``RandBLAS::util::sample_indices_iid_uniform``.

rileyjmurray:
> @rileyjmurray I'm getting a bit of a weird output that the hypergeometric_value returns nan. For example, with a simple d=9, k=7, observed_k=5, I get the following:
> 
> ± |main U:1 ✗| → ./fisher_test Empirical CDF: 0 0 0 0 0 0.4375 0.96875 1 Hypergeometric PMF for d=9, k=7, observed_k=5: nan
> 
> I am calling it like this: `double pmf_test = RandBLAS_StatTests::hypergeometric_pmf(d, d, k, 5);` `std::cout << "Hypergeometric PMF for d=" << d << ", k=" << k << ", observed_k=5: " << pmf_test << std::endl;`
> 
> I added my test project files so you should be able to see everything.

There's a chance that my hypergeometric_pmf function had a bug. You can check by running the code under a debugger, like gdb (for linux) or lldb (for macos). In order to enable debugging you need the CMake configuration line to include ``-DCMAKE_BUILD_TYPE=Debug`` or edit CMakeCache.txt to set BUILD_TYPE=Debug.

RylieWeaver:
Ah, this makes sense. I misread the comment on hypergeometric_pmf. A couple questions:

(1) I'm assuming randblas_require() is the usual preferred method of raising errors?
(2) Are you suggesting to replace some randblas_require() statements in hypergeometric_pmf() with return 0.0 instead? Otherwise, I'm not sure when the probability would equal zero.

Thanks!

rileyjmurray:
> (1) I'm assuming randblas_require() is the usual preferred method of raising errors?

For now, yes. In the future I'd like to revisit this approach, but it's definitely what we do now.

> (2) Are you suggesting to replace some randblas_require() statements in hypergeometric_pmf() with return 0.0 instead? Otherwise, I'm not sure when the probability would equal zero.

Returning 0 is probably appropriate in most situations. However, if someone provides input that makes absolutely no sense (like ``N`` being negative) then an error might be a good signal to the user that they did something wrong. I'll defer the specifics to you.

RylieWeaver:
> For now, yes. In the future I'd like to revisit this approach, but it's definitely what we do now.
Cool

> Returning 0 is probably appropriate in most situations. However, if someone provides input that makes absolutely no sense (like N being negative) then an error might be a good signal to the user that they did something wrong. I'll defer the specifics to you.
In my mind, anything that would return probability=0 makes no sense, and would be more helpful to throw an error than to let it slide by. If you feel differently abut a certain case let me know and I can change it.

RylieWeaver:
Here's an updates with a notable changes and a question:


Notable Change:
- The possible range for observed_k is not simply just 0<=K and this required some logic to be added in certain areas.


Questions
- I'm somewhat stumped on the following error and suspect it has to do with the theoretical CDF not summing to 1. Do you have any thoughts? If it's not quickly obvious, I can dive into debugging, but I wanted to ask in case 5-10 minutes for you could save 1-2 hours for me.
```
Empirical CDF: 0 0 0 0 0.5 0.7 1 
N: 14 K: 8 D: 8
Observed k: 2
N: 14 K: 8 D: 8
Observed k: 3
N: 14 K: 8 D: 8
Observed k: 4
N: 14 K: 8 D: 8
Observed k: 5
N: 14 K: 8 D: 8
Observed k: 6
N: 14 K: 8 D: 8
Observed k: 7
N: 14 K: 8 D: 8
Observed k: 8
Theoretical CDF: 0 0 0.00932401 0.121212 0.470862 0.843823 0.983683 
Significance level: 0.05
CDF1: 0 0 0.00932401 0.121212 0.470862 0.843823 0.983683 
CDF2: 0 0 0 0 0.5 0.7 1 
Diff: 0 CritVal: 0.327333
Diff: 0 CritVal: 0.327333
Diff: 0.00932401 CritVal: 0.327333
Diff: 0.121212 CritVal: 0.327333
Diff: 0.0291375 CritVal: 0.327333
Diff: 0.143823 CritVal: 0.327333
Diff: 0.016317 CritVal: 0.327333

LINE BEFORE ERROR
fisher_test(73241,0x1ff380c00) malloc: Incorrect checksum for freed object 0x152e09f60: probably modified after being freed.
Corrupt value: 0x3f35d2d18a2fe8ca
fisher_test(73241,0x1ff380c00) malloc: *** set a breakpoint in malloc_error_break to debug
Abort trap: 6
```

rileyjmurray:
> * I'm somewhat stumped on the following error and suspect it has to do with the theoretical CDF not summing to 1. Do you have any thoughts? If it's not quickly obvious, I can dive into debugging, but I wanted to ask in case 5-10 minutes for you could save 1-2 hours for me.

Unfortunately I don't see an obvious cause. I think it's reasonable to focus on figuring out why the PMF isn't summing to 1. That problem has the advantage of you being able to verify correctness of your PMF function by testing it with simple cases. Basically -- you can write purely deterministic tests to verify that your statistical-test-related helper functions are actually doing what you want.

If you continue to see this error with malloc then you can compile your code with the address sanitizer turned on (add ``-DSANITIZE_ADDRESS=ON`` to the CMake configuration line, like https://github.com/BallisticLA/RandBLAS/blob/2bebe4a89e16f6110e4513222a49c48b5d65faa1/.github/workflows/core-linux.yaml#L61). You should also do this debugging with compiler optimizations turned off (change -O3 to -O0).

rileyjmurray:
Any updates, @RylieWeaver? I have a PR open that I'd like to merge in the near future. It would be good for us to merge your PR first, since it's much simpler.

RylieWeaver:
> Any updates, @RylieWeaver? I have a PR open that I'd like to merge in the near future. It would be good for us to merge your PR first, since it's much simpler.

Unfortunately, I'm still stuck on this bug. I will be giving it a good chunk of time this Saturday/Sunday. If that's too late, I can clear this up to only add things that are not bugged and merge tonight.

RylieWeaver:
@rileyjmurray I found the bug and got some practice in the C++ debugger along the way. I think all I need to know now is:

- Where is a good place to add the functions relating specifically to fisher-yates? Would it be a new file like test/test_basic_rng/test_shuffle.cc ?


Also, notable change:

- I decided to define the pmf/cdf along 0-->K. This makes it easier for the caller to index the arrays in debugging and did end up requiring some probability=0 logic as you said.

rileyjmurray:
That's great to hear, @RylieWeaver! Regarding your question
>
> Where is a good place to add the functions relating specifically to fisher-yates? Would it be a new file like test/test_basic_rng/test_shuffle.cc ?
> 

I'd put them in ``test_sample_indices.cc``. That file currently only has sampling with replacement, but I intended for it to contain tests for sampling without replacement as well.

RylieWeaver:
@rileyjmurray Ok, great! I reworked some things so that the additions "fit" with the current formatting and usage of the file. This included using pre-built functions when possible and adding some tests. When I rebuilt and ran 'ctest', I got this output, which is promising:

`
278/283 Test #278: TestSampleIndices.iid_ks_skeptical ...............................   Passed    0.12 sec
        Start 279: TestSampleIndices.fisher_yates_ks_generous
279/283 Test #279: TestSampleIndices.fisher_yates_ks_generous .......................   Passed    2.02 sec
        Start 280: TestSampleIndices.fisher_yates_ks_moderate
280/283 Test #280: TestSampleIndices.fisher_yates_ks_moderate .......................   Passed    1.96 sec
        Start 281: TestSampleIndices.fisher_yates_ks_skeptical
281/283 Test #281: TestSampleIndices.fisher_yates_ks_skeptical ......................   Passed    1.98 sec
        Start 282: TestRandom123KnownAnswers.base_generators
282/283 Test #282: TestRandom123KnownAnswers.base_generators ........................   Passed    0.01 sec
        Start 283: TestRandom123KnownAnswers.uniform_histograms
283/283 Test #283: TestRandom123KnownAnswers.uniform_histograms .....................   Passed    0.01 sec

100% tests passed, 0 tests failed out of 283
`

Notable Changes:
-  I made a special_increment_k function so that the repeated_fisher_yates scaled better for testing. The problem arises because a fisher-yates test can be done for any K<=N, and so an exhaustive search is expensive. It's possible that we only want to try certain k and this isn't necessary, but I am not sure.

Upcoming?:
- I'm still getting good with C++ classes, but I feel like these tests could be an optimal opportunity for inheritance. The layout could look like:

### rng_common.hh:    
- `**class KolmogorovSmirnovTester**`: 
- check_critval()
- run_test()

- Custom Functions:
- true_pdf()
- true_cdf()
- observed_pdf()
- observed_cdf()
- kolmogorov_smirnov_tester()
-

rileyjmurray:
This is great, @RylieWeaver! Regarding
> I made a special_increment_k function so that the repeated_fisher_yates scaled better for testing. The problem arises because a fisher-yates test can be done for any K<=N, and so an exhaustive search is expensive. It's possible that we only want to try certain k and this isn't necessary, but I am not sure.

It's indeed the case that exhaustive search is unnecessary. In practice we care about this functionality when $1 \leq K \leq 10$ and $500 \leq N \leq 10{,}000$. It's fine if our tests don't focus on this specific range of values. After all, it should be possible to establish mathematical correctness of the code at much smaller scales.

My one criticism with the current range of values is that having Fisher-Yates tests run for 6 seconds seems excessive. When I run RandBLAS tests on my laptop they take ~4 seconds in total. Of course, we have different machines, so times aren't directly comparable. Let's say that Fisher-Yates tests shouldn't take more than 0.5x the total time of all other tests (so if the original test suite runs in 10 seconds on your machine then we'd want Fisher-Yates to take no longer than 5 seconds). 

Other than that, I have a minor style comment regarding placement of braces, ``{, }``. Here's an example of the style used in RandBLAS:
```c++
        if (K <= K_bounds[0]) {
            // Step one-by-one up to K_bounds[0] (e.g., 9)
            return K + 1;
        } else if (K <= K_bounds[1]) {
            // Step in square root scale up to K_bounds[1] (e.g., 99)
            return static_cast<int64_t>(std::pow(std::sqrt(K) + 1, 2));
        } else {
            // Step in log-scale after K_bounds[1]
            return static_cast<int64_t>(K * exp_base);
        }
```

As for designing test classes with inheritance, let's defer that to a second PR. I'd like to get these changes merged in ASAP, so that I can merge the other open PR right after. Once both PRs are merged it'd be great if you could propose a class-based approach to clean up (part of) the statistical tests. How's that sound?

RylieWeaver:
Ok sounds good @rileyjmurray.

(1) I cut the test a little and reformatted. It should hopefully be good on both now.
(2) Yep. I was unclear in the suggestion, but I was also thinking of this being a future PR, not current.

Let me know if there is anything else to clean up before merging!

rileyjmurray:
Glad we're on the same page for scope of this PR. 

You've probably been notified by GitHub that I just merged. Thanks for the great work! I'll ping you once I merge the other pending PR, at which point you'd be free to start on another project (as you're so inclined).

Issue #99, opened 2024-06-26
Contributor and style guide


rileyjmurray:
See https://github.com/BallisticLA/RandBLAS/blob/main/STYLE_GUIDE.md for our style guide.

rileyjmurray:
Note to self: see https://github.com/tlapack/tlapack/blob/master/CONTRIBUTING.md for a nice model.

rileyjmurray:
Marking as closed by a sequence of commits on Aug 16, 2026. Original comment as follows.

----------------------------

Our style guide is actively maintained. You can find it at:
https://github.com/BallisticLA/RandBLAS/blob/main/STYLE_GUIDE.md.


We need a contributor/style guide. See below for a starting point.

### Available documentation

* We have several files called ``DevNotes.md`` that contain notes for contributors/developers. These might not be totally up to date as of Feb 15, 2025.
* A lot of work went into RandBLAS readthedocs: https://randblas.readthedocs.io/en/stable/index.html. The contributors would benefit from reading the [FAQ](https://randblas.readthedocs.io/en/stable/FAQ.html) in addition to the [tutorial](https://randblas.readthedocs.io/en/stable/tutorial/index.html) and the [API reference](https://randblas.readthedocs.io/en/stable/api_reference/index.html).
* Pages 17 to 28 of the [final report](https://www.osti.gov/servlets/purl/2462906) of a 2024 SNL LDRD project on RandBLAS.

### What C++ is and isn't on the table

Use of the c++ standard library in RandBLAS is subject to two constraints.
1. No use of std::random anywhere, except possibly in our test suite. All random number generation (outside select cases in testing infrastructure) needs to happen with Random123 and be mediated by the RandBLAS-defined ``RNGState'' type.
2. No c++ data structures appear in the public API. So, no std::vector in the signatures of public functions. The public API is defined as what we put in the web documentation.

### Dev infrastructure

#### Don't rely on Homebrew. Use spack.
Homebrew bit me in the butt today, so I bit the bullet and downloaded Spack. It's a lifesaver! Here's how you'd "install" in bash:
```bash
export HOMEDIR="~"
# ^ Ideally you'd use a full-qualified path, like /users/riley/.
cd $HOMEDIR
git clone https://github.com/spack/spack.git
# ^ you'll now have a spack folder in your home directory.
printf "\nif [ -f ~/.bashrc ]; then \n\tsource ~/.bashrc\n\n" >> .bash_profile
[...]

PR #122, opened 2024-10-01
[WiP] SRHTs

aryamanjeendgar:
This is a draft PR for integrating SRHTs into `RandBLAS`

### DIFF
--- a/RandBLAS/trig_skops.hh
+++ b/RandBLAS/trig_skops.hh
@@ -1,0 +1,318 @@
+#include "RandBLAS/base.hh"
+#include "RandBLAS/exceptions.hh"
+#include "RandBLAS/random_gen.hh"
+#include <RandBLAS/sparse_skops.hh>
+
+#include <Random123/philox.h>
+#include <blas.hh>
+#include <lapack.hh>
+#include <omp.h>
+
+#include <iostream>
+#include <stdio.h>
+#include <stdexcept>
+#include <string>
+#include <tuple>
+#include <random>
+
+#include <math.h>
+#include <typeinfo>
+
+#define MAX(a, b) (((a) < (b)) ? (b) : (a))
+#define MIN(a, b) (((a) < (b)) ? (a) : (b))
+
+namespace RandBLAS {
+
+    // =============================================================================
+    /// WARNING: None of the following functions or overloads thereof are part of the
+    /// public API
+    ///
+    void generateRademacherVector_r123(std::vector<int64_t> &buff, uint32_t key_seed, uint32_t ctr_seed, int n) {
+        typedef r123::Philox2x32 RNG;
+        // std::vector<int64_t> rademacherVector(n);
+
+        // Use OpenMP to parallelize the Rademacher vector generation
+        #pragma omp parallel
+        {
+            // Each thread has its own RNG instance
+            RNG rng;
+
+            // You can use any seed value for the key
+            RNG::key_type k = {{key_seed + omp_get_thread_num()}}; // Unique key for each thread
+
+            // Thread-local counter
+            RNG::ctr_type c;
+
+            // Parallel for loop
+            #pragma omp for
+            for (int i = 0; i < n; ++i) {
+                // Set the counter for each random number (unique per thread)
+                c[0] = ctr_seed + i; // Ensure the counter is unique across threads
+
+                // Generate a 2x32-bit random number using the Philox generator
+                RNG::ctr_type r = rng(c, k);
+
+                // Convert the random number into a float in [0, 1) using u01fixedpt
+                float randValue = r123::u01fixedpt<float>(r.v[0]);
+
+                // Convert the float into a Rademacher entry (-1 or 1)
+                buff[i] = randValue < 0.5 ? -1 : 1;
+            }
+        }
+    }
+
+    std::vector<int64_t> generateRademacherVector_parallel(int64_t n) {
+        std::vector<int64_t> rademacherVec(n);
+
+        #pragma omp parallel
+        {
+            // Each thread gets its own random number generator
+            std::random_device rd;
+            std::mt19937 gen(rd());
+            std::bernoulli_distribution bernoulli(0.5);
+
+            #pragma omp for
+            for (int64_t i = 0; i < n; ++i) {
+                rademacherVec[i] = bernoulli(gen) ? 1 : -1;
+            }
+        }
+
+        return rademacherVec;
+    }
+
+    template<typename T>
+    void applyDiagonalRademacher(int64_t rows, int64_t cols, T* A) {
+        std::vector<int64_t> diag = generateRademacherVector_parallel(cols);
+
+        for(int col=0; col < cols; col++) {
+        if(diag[col] > 0)
+            continue;
+        blas::scal(rows, diag[col], &A[col * rows], 1);
+    }
+    }
+
+    template<typename T>
+    void applyDiagonalRademacher(int64_t rows, int64_t cols, T* A, std::vector<int64_t> &diag) {
+        //NOTE: Only considers sketching from the left + ColMajor format as of now
+        for(int col=0; col < cols; col++) {
+        if(diag[col] > 0)
+            continue;
+        blas::scal(rows, diag[col], &A[col * rows], 1);
+    }
+    }
+
+    // `applyDiagonalRademacher`should have a similar API as `sketch_general`
+    // Will be called inside of `lskget`
+    // B <- alpha * diag(Rademacher) * op(A) + beta * B
+    // `alpha`, `beta` aren't needed and shape(A) == shape(B)
+    // lda == ldb
+    template<typename T>
+    void applyDiagonalRademacher(
+        blas::Layout layout,
+        blas::Op opA, // taken from `sketch_general/lskget`
+        int64_t n, // everything is the same size as `op(A)`: m-by-n
+        int64_t m,
+        std::vector<int64_t> &diag,
+        const T* A, // The data matrix that won't be modified
+        // int64_t lda,
+        T* B // The destination matrix
+        // int64_t ldb
+    ) {
+        // B <- alpha * diag(Rademacher) * op(A) + beta * B
+        // `alpha`, `beta` aren't needed
+        // && shape(A) == shape(B) && lda == ldb && shape({A | B}) = m-by-n
+        //NOTE: Use `layout` and `opA` for working with RowMajor data
+        int64_t lda = m;
+        int64_t ldb = m;
+
+        //NOTE: Should the copy be made inside of `applyDiagonalRademacher` or
+        // should it be inside of `lskget`?
+        lapack::lacpy(lapack::MatrixType::General, m, n, A, lda, B, ldb);
+
+        applyDiagonalRademacher(m, n, B, diag);
+    }
+
+
+    // `permuteRowsToTop` should just take in `B` as an input
+    // `B` will be passed in by the user to `sketch_general/lskget` and `permuteRowsToTop` will take in
+    // the `B` that has been modified by `applyRademacherDiagonal`
+    // Will also be called inside of `lskget`
+    template<typename T>
+    void permuteRowsToTop(int rows, int cols, std::vector<int64_t> selectedRows, T* B, int ldb) {
+        //NOTE: There should be a similar `permuteColsToLeft`
+        int top = 0;  // Keeps track of the topmost unselected row
+
+        // Compute `r` internally
+        // int64_t r = 100;
+        // std::vector<int64_t> selectedRows = generateRademacherVector_parallel(r);
+
+        for (int selected : selectedRows) {
+            if (selected != top) {
+                // Use BLAS swap to swap the entire rows
+                // Swapping row 'selected' with row 'top'
+                blas::swap(cols, &B[selected], ldb, &B[top], ldb);
+            }
+            top++;
+        }
+    }
+
+    enum class TrigDistName: char {
+        Fourier = 'F',
+
+        // ---------------------------------------------------------------------------
+
+        Hadamard = 'H'
+    };
+
+    struct TrigDist {
+        const int64_t n_rows;
+        const int64_t n_cols;
+
+        int64_t dim_short;
+        int64_t dim_long;
+
+        const TrigDistName family;
+
+        TrigDist(
+            int64_t n_rows,
+            int64_t n_cols,
+            TrigDistName tn = TrigDistName::Hadamard
+        ) : n_rows(n_rows), n_cols(n_cols), family(tn) {
+            dim_short = MIN(n_rows, n_cols);
+            dim_long = MAX(n_rows, n_cols);
+}
+};
+
+    template<typename T, SignedInteger sint_t, typename RNG = r123::Philox4x32>
+    struct TrigSkOp {
+        using generator = RNG;
+        using state_type = RNGState<RNG>;
+        using buffer_type = T;
+
+        const int64_t n_rows;
+        const int64_t n_cols;
+        int64_t dim_short;
+        int64_t dim_long;
+        // int64_t n_sampled;
+
+        const TrigDist dist;
+
+
+        const RNGState<RNG> seed_state;
+
+        RNGState<RNG> next_state;
+
+        const blas::Layout layout;
+        const bool sketchFromLeft = true;
+        bool known_filled = false;
+
+        T* DiagonalRademacher = nullptr;
+        T* SampledRows = nullptr;
+
+        TrigSkOp(
+            TrigDist dist,
+            RNGState<RNG> const &state,
+            bool known_filled = true
+        ) : n_rows(dist.n_rows), n_cols(dist.n_cols), dist(dist), seed_state(state), known_filled(known_filled), dim_short(dist.dim_short), dim_long(dist.dim_long){
+            // Memory for Rademacher diagonal gets allocated here
+            // Number of rows/cols to be sampled gets decided here
+            // i.e. `n_sampled` gets set
+            if(sketchFromLeft)
+                DiagonalRademacher = new T[n_rows];
+            else
+                DiagonalRademacher = new T[n_cols];
+
+            SampledRows = new sint_t[dim_short];
+
+            //TODO: Logic to compute the number of samples that we require `r`
+            // this can be shown to depend on the maximal row norm of the data matrix
+            //NOTE: Do not have access to this in the data-oblivious regime --- how
+            // do we get access?
+};
+
+        TrigSkOp(
+            TrigDistName family,
+            int64_t n_rows,
+            int64_t n_cols,
+            uint32_t key
+        ) : n_rows(n_rows), n_cols(n_cols), dist(TrigDist{n_rows, n_cols, family}), seed_state() {};
+
+        ~TrigSkOp();
+};
+
+// Populates the Rademacher diagonal entries and decides the rows/cols to
+// be sampled at the end
+template<typename T, typename RNG, SignedInteger sint_t>
+RandBLAS::RNGState<RNG> fill_trig(
+    TrigSkOp<T, RNG> &Tr
+) {
+    /**
+     * Will do the work of filling in the diagonal Rademacher entries
+     * and selecting the rows/cols to be sampled
+     */
+
+    auto [ctr, key] = Tr.seed_state;
+
+    // Fill in the Rademacher diagonal
+    if(Tr.sketchFromLeft)
+        generateRademacherVector_r123(Tr.DiagonalRademacher, key, ctr, Tr.n_rows);
+    else
+        generateRademacherVector_r123(Tr.DiagonalRademacher, key, ctr, Tr.n_cols);
+
+    //NOTE: Select the rows/cols to be sampled --- use the `repeated_fisher_yates` function
+    int64_t r = Tr.dim_short;
+    int64_t d = Tr.dim_long;
+
+    // Output containers
+    std::vector<sint_t> idxs_major(r); // Stores the sampled integers
+    std::vector<sint_t> idxs_minor(r); // Not needed in your case
+    std::vector<T> vals(r); // Not needed in your case, but required by the function
+
+    RandBLAS::repeated_fisher_yates<T, RNG, sint_t>(
+        Tr.seed_state,
+        r,         // Number of samples (vec_nnz)
+        d,         // Total number of elements (dim_major)
+        1,         // Single sample round (dim_minor)
+        Tr.SampledRows,  // Holds the required output
+        idxs_minor.data(),  // Placeholder
+        vals.data()         // Placeholder
+    );
+}
+}
+
+
+namespace RandBLAS::trig {
+
+// Performs the actual application of the fast trigonometric transform
+// Only called after calling `fill_trig`
+template <typename T, typename RNG>
+inline void lskget(
+    blas::Layout layout,
+    blas::Op opS,
+    blas::Op opA,
+    int64_t d, // B is d-by-n
+    int64_t n, // op(A) is m-by-n
+    int64_t m, // op(S) is d-by-m
+    T alpha,
+    TrigSkOp<T, RNG> &Tr,
+    int64_t ro_s,
+    int64_t co_s,
+    const T* A, // data-matrix
+    int64_t lda,
+    T beta,
+    T* B, // output matrix
+    int64_t ldb
+) {
+    if (!Tr.known_filled)
+        fill_trig(Tr);
+
+    // Applying the diagonal transformation
+    applyDiagonalRademacher(layout, opA, n, m, Tr.DiagonalRademacher, A, B);
+    // applyDiagonalRademacher(m, n, A, Tr.DiagonalRademacher);
+
+    //TODO: Apply the Hadamard transform
+
+    //... and finally permute the rows
+    permuteRowsToTop(m, n, Tr.SampledRows, B, ldb);
+}
+}


RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,315 @@
...
+    template<typename T, SignedInteger sint_t, typename RNG = r123::Philox4x32>
+    struct TrigSkOp {
+        using generator = RNG;
+        using state_type = RNGState<RNG>;
+        using buffer_type = T;
+
+        const int64_t n_rows;
+        const int64_t n_cols;
+        int64_t dim_short;
+        int64_t dim_long;
```
rileyjmurray, line 194: Same comment as before; these should be const.

Please make sure the initialization order of the various members follows the same pattern as SparseSkOp and DenseSkOp.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,315 @@
...
+
+        const TrigDist dist;
+
+
+        const RNGState<RNG> seed_state;
+
+        RNGState<RNG> next_state;
+
+        const blas::Layout layout;
+        const bool sketchFromLeft = true;
```
rileyjmurray, line 205: I'll trust that you'll remove this soon.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,315 @@
...
+        bool known_filled = false;
+
+        T* DiagonalRademacher = nullptr;
+        T* SampledRows = nullptr;
```
rileyjmurray, line 209: 1. Use snake case, not camel case.
2. sampled rows should be ``int64_t*``, not ``T*``.
3. rename "sampled rows" to something like ``subsample_idxs`` or ``sampled_coords`` or ``subsampled_coords``. This way it can refer to either rows or columns.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,315 @@
...
+            RNGState<RNG> const &state,
+            bool known_filled = true
+        ) : n_rows(dist.n_rows), n_cols(dist.n_cols), dist(dist), seed_state(state), known_filled(known_filled), dim_short(dist.dim_short), dim_long(dist.dim_long){
+            // Memory for Rademacher diagonal gets allocated here
+            // Number of rows/cols to be sampled gets decided here
+            // i.e. `n_sampled` gets set
+            if(sketchFromLeft)
+                DiagonalRademacher = new T[n_rows];
+            else
+                DiagonalRademacher = new T[n_cols];
```
rileyjmurray, line 222: Fun fact: the correct length of this vector is always ``dim_long``; you don't need to think about whether you're sketching from the left or right.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,315 @@
...
+
+            SampledRows = new sint_t[dim_short];
+
+            //TODO: Logic to compute the number of samples that we require `r`
+            // this can be shown to depend on the maximal row norm of the data matrix
+            //NOTE: Do not have access to this in the data-oblivious regime --- how
+            // do we get access?
```
rileyjmurray, line 229: This is outside RandBLAS' scope for now. It's the kind of thing we can have in our web documentation.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,315 @@
...
+};
+
+        TrigSkOp(
+            TrigDistName family,
+            int64_t n_rows,
+            int64_t n_cols,
+            uint32_t key
+        ) : n_rows(n_rows), n_cols(n_cols), dist(TrigDist{n_rows, n_cols, family}), seed_state() {};
+
+        ~TrigSkOp();
```
rileyjmurray, line 239: The destructor needs to call ``delete []`` on the arrays you allocated with ``new``.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,315 @@
...
+     * and selecting the rows/cols to be sampled
+     */
+
+    auto [ctr, key] = Tr.seed_state;
+
+    // Fill in the Rademacher diagonal
+    if(Tr.sketchFromLeft)
+        generateRademacherVector_r123(Tr.DiagonalRademacher, key, ctr, Tr.n_rows);
+    else
+        generateRademacherVector_r123(Tr.DiagonalRademacher, key, ctr, Tr.n_cols);
```
rileyjmurray, line 259: 1. it suffices to use ``generateRademacherVector_r123(Tr.DiagonalRademacher, key, ctr, Tr.dim_long);``
2. the generateRademacherVector_r123 function needs to return an updated counter.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,315 @@
...
+
+    //NOTE: Select the rows/cols to be sampled --- use the `repeated_fisher_yates` function
+    int64_t r = Tr.dim_short;
+    int64_t d = Tr.dim_long;
```
rileyjmurray, line 263: I'd advise against defining these sorts of temporary variables. Just substitute ``Tr.dim_short`` and ``Tr.dim_long``.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,315 @@
...
+        ) : n_rows(dist.n_rows), n_cols(dist.n_cols), dist(dist), seed_state(state), known_filled(known_filled), dim_short(dist.dim_short), dim_long(dist.dim_long){
+            // Memory for Rademacher diagonal gets allocated here
+            // Number of rows/cols to be sampled gets decided here
+            // i.e. `n_sampled` gets set
+            if(sketchFromLeft)
+                DiagonalRademacher = new T[n_rows];
+            else
+                DiagonalRademacher = new T[n_cols];
+
+            SampledRows = new sint_t[dim_short];
```
rileyjmurray, line 224: Yeah! This is the right idea for using dim_short.

rileyjmurray (commented):
It's exciting to see progress here!

I left several comments. Please resolve them, but then take a step back. Part of what you're doing is adding FHT support. This already takes effort without thinking about randomization. 

### Building up your suite of FHT kernels
Make a folder ``RandBLAS/trig/`` and a file ``RandBLAS/trig/hadamard.hh``. Have your various FHT implementations go here. You already have one for applying to column-major data from the left and storing the result in a column-major sketch. You could easily write implementation that takes in row-major data and writes to a row-major sketch (although the only way you can parallelize this is with BLAS1, I think).

It would be great to have implementations for both the Hadamard transform itself and for the transpose of the Hadamard transform (equivalently, the inverse of the Hadamard transform). If you do this then w.l.o.g. you can always assume you're applying the transformation from the left.

Once you have those functions working, add in the ability to implicitly scale the rows of the input matrix by a vector of coefficients. In sketching we end up setting this vector to a Rademacher random vector, but from an implementation standpoint these functions don't need to care where the vector comes from.

Once you have those cases sorted out, you can allow conflicting layouts for the input matrix and the output matrix (i.e., input is row-major and output is column-major). This is the trick I use for resolving transposition in the sparse-times-dense matrix kernels.

Note: if you feel like you need to allocate a temporary matrix for workspace in order to do anything useful, you can definitely try that.

### Writing tests
Make a folder ``test/test_matmul_cores/test_trig/`` and a file ``test/test_matmul_cores/test_trig/test_hadamard.cc``. This will handle tests only for your FHT. 

You can take some inspiration from https://github.com/BallisticLA/RandBLAS/blob/36da117dd3fee22fa240722efde359651b9711c7/test/test_matmul_wrappers/test_sketch_sparse.cc#L53-L190 to see what your tests could like. But don't let this line of thinking hobble you. Feel free to take a different approach to get started. I can give feedback and we iterate a bit.

rileyjmurray:
@aryamanjeendgar, for our reference, here's the code you mentioned from the test suite of FFHT:
```c
        void fht(double *buf, int log_n) {
        int n = 1 << log_n;
        for (int i = 0; i < log_n; ++i) {
                int s1 = 1 << i;
                int s2 = s1 << 1;
                for (int j = 0; j < n; j += s2) {
                for (int k = 0; k < s1; ++k) {
                        double u = buf[j + k];
                        double v = buf[j + k + s1];
                        buf[j + k] = u + v;
                        buf[j + k + s1] = u - v;
                }
                }
        }
        }
```
Step 1 to figuring out how we'll change it: replace the bit manipulations with standard integer arithmetic (or just give explanatory comments).

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,493 @@
...
+    void generateRademacherVector_r123(sint_t* buff, uint32_t key_seed, uint32_t ctr_seed, int64_t n) {
+        typedef r123::Philox2x32 RNG;
+
+        // Use OpenMP to parallelize the Rademacher vector generation
+        #pragma omp parallel
+        {
+            // Each thread has its own RNG instance
+            RNG rng;
+
+            RNG::key_type k = {{key_seed + omp_get_thread_num()}}; // Unique key for each thread
```
rileyjmurray, line 40: 1. This function should take in an RNGState and return an RNGState.
2. Parallel generation should not rely on different keys. In the situations where we need parallelism we make a point to ensure that the number of threads used doesn't affect the outcome of the random number generation.
3. I'm doubtful that parallelism is useful here. We use parallelism when generating many vectors at once. I have a preference for _no OpenMP here_ unless you can show there's a compelling performance reason to keep it.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,493 @@
+#include "RandBLAS/base.hh"
+#include "RandBLAS/exceptions.hh"
+#include "RandBLAS/random_gen.hh"
+#include <RandBLAS/sparse_skops.hh>
+
+#include <Random123/philox.h>
+#include <blas.hh>
+#include <lapack.hh>
+#include <omp.h>
```
rileyjmurray, line 9: RandBLAS doesn't have OpenMP as a hard dependency. So you need any ``include <omp.h>`` and any OpenMP function calls to be wrapped with compiler preprocessor directives.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,493 @@
...
+
+#include <iostream>
+#include <stdio.h>
+#include <stdexcept>
+#include <string>
+#include <tuple>
+#include <random>
```
rileyjmurray, line 16: The only place that std::random is allowed in RandBLAS is in our test code. Remove this include, any functions that rely on it.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,493 @@
...
+
+namespace RandBLAS {
+
+    // =============================================================================
+    /// WARNING: None of the following functions or overloads thereof are part of the
+    /// public API
+    ///
+    template<SignedInteger sint_t>
+    void generateRademacherVector_r123(sint_t* buff, uint32_t key_seed, uint32_t ctr_seed, int64_t n) {
+        typedef r123::Philox2x32 RNG;
```
rileyjmurray, line 32: Random number generators should be template parameters, and use
```
template <typename RNG = DefaultRNG, ...>
```
where ``DefaultRNG = r123::Philox4x32`` is defined in RandBLAS/base.hh.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,493 @@
...
+            }
+        }
+        else {
+            for(int row = 0; row < rows; row++) {
+                if(diag[row] > 0)
+                    continue;
+                blas::scal(cols, diag[row], &A[row * cols], 1);
+            }
+        }
+    }
```
rileyjmurray, line 131: This is fine for now, but it can and should be much more efficient. It's also probably better suited to RandBLAS/util.hh.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,493 @@
...
+        //NOTE: Use `layout` and `opA` for working with RowMajor data
+        int64_t lda = m;
+        int64_t ldb = m;
+
+        //NOTE: Should the copy be made inside of `applyDiagonalRademacher` or
+        // should it be inside of `lskget`?
+        lapack::lacpy(lapack::MatrixType::General, m, n, A, lda, B, ldb);
+
+        applyDiagonalRademacher(m, n, B, diag);
+    }
```
rileyjmurray, line 163: What's this for? We discussed that the functions you're working on need to do everything in-place.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,493 @@
...
+            for(int row = 0; row < rows; row++) {
+                if(diag[row] > 0)
+                    continue;
+                blas::scal(cols, diag[row], &A[row * cols], 1);
+            }
+        }
+    }
+
+    // `applyDiagonalRademacher`should have a similar API as `sketch_general`
+    // Will be called inside of `lskget`
```
rileyjmurray, line 134: Update comments to be consistent with the project plan.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,493 @@
...
+    void permuteRowsToTop(
+                          blas::Layout layout,
+                          int64_t rows,
+                          int64_t cols,
+                          sint_t* selectedRows,
+                          int64_t d, // size of `selectedRows`
+                          T* A
+                          ) {
+        //TODO: There should be a similar `permuteColsToLeft` for sketching
+        //from the right
```
rileyjmurray, line 180: Good observation!

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,493 @@
...
+                        else if(!b2 && b1) {
+                            col_buf[j + k] = u + v;
+                        }
+                        else if(!b2 && !b1)
+                            continue;
+                    }
+                }
+            }
+        }
+    }
```
rileyjmurray, line 269: 1. Numerical precision should be templated.

rileyjmurray (changes requested):
Please do a detailed read of my notes in GitHub Issue #99.

I didn't review the whole PR since my plane is landing and I need to put away my laptop.

RandBLAS/trig_skops.hh
aryamanjeendgar: I didn't end up using OpenMP anywhere else (so-far), so I've dropped the dependency for now!

RandBLAS/trig_skops.hh
aryamanjeendgar: Added an overload of the `generateRademacherVector_r123` method with the appropriate signature

RandBLAS/trig_skops.hh
aryamanjeendgar: Left it in by accident, removed!

RandBLAS/trig_skops.hh
aryamanjeendgar: Added an overload of the generateRademacherVector_r123 method with the appropriate signature and removed the unnecessary OpenMP dependency (the performance gain in the case of `std::random` was very significant when switching to a parallel implementation but not so much in the case of Random123)

RandBLAS/trig_skops.hh
aryamanjeendgar: Both `permuteColsToTop` and `permuteColsToLeft` have been implemented!

RandBLAS/trig_skops.hh
aryamanjeendgar: Made input buffers templated!

RandBLAS/trig_skops.hh
aryamanjeendgar: Resolved with the latest commit!

RandBLAS/trig_skops.hh
aryamanjeendgar: Using snake_case uniformly throughout all of my code now!

rileyjmurray (commented):
This is looking pretty good! Please start writing unit tests.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,440 @@
...
+                        if(b1 && b2) {
+                            col_buf[j + k] = u + v;
+                            col_buf[j + k + s1] = u - v;
+                        }
+                        else if(!b2 && b1) {
+                            col_buf[j + k] = u + v;
+                        }
+                        else if(!b2 && !b1)
+                            continue;
+                    }
```
rileyjmurray, line 209: @aryamanjeendgar, sorry, jumped the gun. Please replace the ``double`` types here with ``T``. You'll need to do this in the other functions as well.

aryamanjeendgar: Thank you for pointing this out! Corrected it!

aryamanjeendgar:
A description of the tests:

All of the tests right now (unfortunately) use Eigen --- I use Eigen extensively to be able to consistently produce `RowMajor` and `ColMajor` random matrices, computing matrix norms and products in all of the tests --- additionally, I use Eigen in the tests for the permutation matrices since Eigen has a convenient out-of-the-box `PermutationMatrix` class.

I also ended up using `std::random` for generating a simple "random" buffer ƒrom which I instantiate the Eigen matrices.

Much of this infrastructure (with norm computation, scaling, computing products etc.) can be ported (with a lot more LoC XD) to BLAS with the exception of the permutation tests which tests against Eigen's `PermutationMatrix` .


The correctness tests are fairly straightforward to understand:
1. For testing permutation: I permute the last row/column of the input matrix (it is swapped with the leading row/column), and then check if the result of my implementation and Eigens' matches up.
2. For testing Hadamard: Checked against explicit application of the Hadamard matrix
3. For testing Diagonal scaling: Scale the input matrix by all $-1$'s via my code and see if it's equal to a negated copy of the initial matrix

The tests are simple, but they cover _all_ of the potential code paths in my code (since the code paths are chosen around two inputs: sketching direction + layout of the input matrix)

Let me know how we should proceed next, @rileyjmurray!

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,440 @@
...
+    // Size of the Rademacher entries = |A_cols|
+    //TODO: Change `diag` to float/doubles (same data type as the matrix)
+    sint_t* diag = new sint_t[n];
+    sint_t* selected_rows = new sint_t[d];
+
+    auto [ctr, key] = random_state;
+
+    //Step 1: Scale with `D`
+        //Populating `diag`
+    generate_rademacher_vector_r123(diag, key[0], ctr[0], n);
```
rileyjmurray, line 365: Pass in the state. Don't pass in the first component of counter and key arrays.

rileyjmurray (changes requested):
0. You don't need the ``RandBLAS::`` namespace qualifications within RandBLAS itself.
1. Don't explicitly reference r123::Philox4x32. Reference RandBLAS::DefaultRNG (But, again, without the namespace qualification).
2. I don't like people having to explicitly pass in ``log_n`` to the various FHT functions. Do you have a good reason for that?
3. Please don't use a single monolithic "correctness" test. 
4. Tests should template numerical precision, not hard-code double. The tolerance for the test would ideally be based on some known error bounds for the operations you're trying to perform. The error bounds you're working with should be very simple since every part of an RHT is an orthogonal matrix, and there are special bounds for those.
5. There should be functions in RandBLAS proper (not just tests) for inverting an RHT.

This is just a partial review. I have to jump to a meeting.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,440 @@
...
+            // Generate a 2x32-bit random number using the Philox generator
+            typename RNG::ctr_type r = rng(c, key);
+
+            // Convert the random number into a float in [0, 1) using u01fixedpt
+            float rand_value = r123::u01fixedpt<float>(r.v[0]);
+
+            // Convert the float into a Rademacher entry (-1 or 1)
+            buff[i] = rand_value < 0.5 ? -1 : 1;
+        }
+    }
```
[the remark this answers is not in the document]
aryamanjeendgar: Done!

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,440 @@
...
+            }
+        }
+        else {
+            for(int64_t row = 0; row < rows; row++) {
+                if(diag[row] > 0)
+                    continue;
+                blas::scal(cols, diag[row], &A[row * cols], 1);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
aryamanjeendgar: That's a much more sensible interpretation! I wrote the code with `left` in the context of the diagonal matrices meaning the matrix is on the left (the hadamard `fht_dispatch` convention is the opposite, which is confusing) --- I've corrected it and `left` is consistent with pre-multiplying the diagonal matrix!

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,440 @@
...
+    template <typename T>
+    void fht_dispatch(
+        bool left,
+        blas::Layout layout,
+        T* buff,
+        int64_t log_n,
+        int64_t num_rows,
+        int64_t num_cols
+        )
+    {
```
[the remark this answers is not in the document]
aryamanjeendgar: Done! (though I do require access to an `lda` argument since the value of `lda` is apparent from the way I deal with the application of the hadamard transform on a case-by-case basis)

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,440 @@
...
+ * These functions apply an in-place, SRHT-like transform to the input matrix
+ * i.e. A <- (\Pi H D)A OR A <- A(D H \Pi) (which is equivalent to A <- A(\Pi H D)^{-1})
+ * layout: Layout of the input matrix (`ColMajor/RowMajor`)
+ * A: (m x n), input dimensions of `A`
+ * d: The number of rows/columns that will be permuted by the action of $\Pi$
+ */
+template <typename T, typename RNG = r123::Philox4x32, SignedInteger sint_t = int64_t>
+inline void lmiget(
+    blas::Layout layout,
+    RandBLAS::RNGState<RNG> random_state,
```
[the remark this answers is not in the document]
aryamanjeendgar: Done!

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,440 @@
...
+namespace RandBLAS::trig {
+/*
+ * These functions apply an in-place, SRHT-like transform to the input matrix
+ * i.e. A <- (\Pi H D)A OR A <- A(D H \Pi) (which is equivalent to A <- A(\Pi H D)^{-1})
+ * layout: Layout of the input matrix (`ColMajor/RowMajor`)
+ * A: (m x n), input dimensions of `A`
+ * d: The number of rows/columns that will be permuted by the action of $\Pi$
+ */
+template <typename T, typename RNG = r123::Philox4x32, SignedInteger sint_t = int64_t>
+inline void lmiget(
```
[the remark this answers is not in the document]
aryamanjeendgar: All `lmiget, rmiget` and `generate_rademacher_vector_r123` now return their updated RNG states

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,440 @@
...
+        //TODO: Do I return this at some point?
+    RandBLAS::RNGState<RNG> next_state = RandBLAS::repeated_fisher_yates<T, RNG, sint_t>(
+        random_state,
+        d,         // Number of samples (vec_nnz)
+        m,         // Total number of elements (dim_major)
+        1,         // Single sample round (dim_minor)
+        selected_rows,  // Holds the required output
+        idxs_minor.data(),  // Placeholder
+        vals.data()         // Placeholder
+    );
```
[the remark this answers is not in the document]
aryamanjeendgar: I'm not sure --- I tried looking after you mentioned, but as-of-now, only this signature shows up when I grep through the RandBLAS codebase --- could you point me to where this overload sits right now?

RandBLAS/trig_skops.hh
aryamanjeendgar: Resolved!

RandBLAS/trig_skops.hh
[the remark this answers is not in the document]
rileyjmurray: https://randblas.readthedocs.io/en/stable/api_reference/utilities.html#_CPPv4I_13SignedInteger0EN8RandBLAS21repeated_fisher_yatesE7state_t7int64_t7int64_t7int64_tP6sint_tRK7state_t

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,440 @@
...
+                          int64_t d, // size of `selected_rows`
+                          T* A
+                          ) {
+        int64_t top = 0;  // Keeps track of the topmost unselected row
+
+        //TODO: discuss precise semantics of `selected_rows` in this function
+        if(layout == blas::Layout::ColMajor) {
+            for (int64_t i=0; i < d; i++) {
+                randblas_error_if_msg(selected_rows[i] == top,
+                                      "The list of provided indices should be unique");
```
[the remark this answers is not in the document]
aryamanjeendgar: I added this when I was trying to adapt a different signature of the permutation functions --- this was a safeguard against the user passing in `0` as one of the indices in their list (because `0` already starts off as a top index by default), but I decided to stick to the signature that I already have in place because it makes a lot more sense from the perspective of the SRHT (I don't see the point of the user having to provide an entire list of indices when they anyways have to just be uniformly sampled by us) --- I will remove this check in the next commit

RandBLAS/trig_skops.hh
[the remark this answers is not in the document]
aryamanjeendgar: Thanks! My branch just wasn't upto-date XD

aryamanjeendgar:
>You don't need the RandBLAS:: namespace qualifications within RandBLAS itself.
Fixed!

>Don't explicitly reference r123::Philox4x32. Reference RandBLAS::DefaultRNG (But, again, without the namespace qualification).

Looks like I didn't have `DefaultRNG` on the previous version that I was working on --- will fix this with the next commit

>I don't like people having to explicitly pass in log_n to the various FHT functions. Do you have a good reason for that?
I use it to implicitly decide the size of the hadamard matrix while applying the transform --- it's not necessary per se, but that's why I made sure to remove the argument from the user-facing functions `lmiget` and `rmiget`

>Please don't use a single monolithic "correctness" test.
I just wanted to make the tests as easy to dispatch as possible --- the test fixtures themselves are a lot more descriptive (like `test_diag_left_rowmajor` --- denoting the module I test, which direction the multiplication happens from and the layout of the test matrix).

>Tests should template numerical precision, not hard-code double. The tolerance for the test would ideally be based on some known error bounds for the operations you're trying to perform. The error bounds you're working with should be very simple since every part of an RHT is an orthogonal matrix, and there are special bounds for those.
With the latest commit all of test matrices etc. I construct are now templated --- since most of the tests involve some kind of a norm check, I just set the tolerance to `1e-5` (but they pass even for `1e-11`)

>There should be functions in RandBLAS proper (not just tests) for inverting an RHT.
This is hard to do without committing to refactor to work with a `TrigSkOp` class rather than just the functions `lmiget, rmiget` --- since to invert the transform, I'd require access to 1) The exact diagonal matrix generated from the `generate_rademacher_vector_r123` and 2) The indices of rows/cols which we permuted (which will need to be reversed for the inversion) --- how do you think we can introduce inversion inside of `lmiget, rmiget`?

rileyjmurray:
> With the latest commit all of test matrices etc. I construct are now templated --- since most of the tests involve some kind of a norm check, I just set the tolerance to 1e-5 (but they pass even for 1e-11)

If you're using a rule of thumb like that then the least you can do is define your rule with respect to the unit roundoff of the current numerical type. In this context it sounds like ``RandBLAS::sqrt_epsilon<T>()`` would do the job.

> [Having proper functions for inverting an RHT] is hard to do without committing to refactor to work with a TrigSkOp class rather than just the functions lmiget, rmiget --- since to invert the transform, I'd require access to 1) The exact diagonal matrix generated from the generate_rademacher_vector_r123 and 2) The indices of rows/cols which we permuted (which will need to be reversed for the inversion) --- how do you think we can introduce inversion inside of lmiget, rmiget?

How about you define a struct type called "HadamardMixingOp" to hold the data needed for lmiget and rmiget. Maybe we end up defining a MixingOperator C++20 concept later. For now you can just do one type.

Given the HadamardMixingOp container, you can define "fill_hadamard" function (or something like that) akin to fill_dense and fill_sparse, which falls back on an implementation of a "fill_hadamard_unpacked_nosub" that's analogous to "fill_sparse_unpacked_nosub."

aryamanjeendgar:
So, right now --- I have written `HadamardMixingOp` to just be a class with all the data that I need on-hand to be able to invert the $\mathbb{\Pi \text{RHT}}$ operation --- with the actual invert  method being a free-function that takes in an instance of `HadamardMixingOp` and the sketched matrix as an input

But the problem is that this instance of `HadamardMixingOp` ought to be constructed inside of `lmiget/rmiget` when the indices for sampling and the rademacher entries are sampled --- and as of right now, since `lmiget/rmiget` are independent of `HadamardMixingOp` --- I perform the memory management for the buffers holding the rademacher entries and the sampled indices inside of the functions themselves

But from what I've realised after thinking about it for a bit, requiring an invert operation would also require us to enforce the creation of a `HadamardMixingOp` instance with every call of `lmiget/rmiget` , for the memory management for all of this to shift inside of `HadamardMixingOp` and most importantly, for `lmiget/rmiget/invert` to become free-methods accepting an instance of `HadamardMixingOp`


i.e. I'm not sure if there is a way to keep the construction of a `HadamardMixingOp` class "optional" should we want the user to have the option of being able to invert the $\mathbb{\Pi \text{RHT}}$

RandBLAS/trig_skops.hh
```diff
@@ -10,47 +10,25 @@
...
-
-            // Convert the float into a Rademacher entry (-1 or 1)
-            buff[i] = rand_value < 0.5 ? -1 : 1;
-        }
-    }
-
-    template<SignedInteger sint_t, typename RNG = r123::Philox4x32>
-    RandBLAS::RNGState<RNG> generate_rademacher_vector_r123(sint_t* buff, int64_t n, RandBLAS::RNGState<RNG> seed_state) {
+    template<SignedInteger sint_t = int64_t, typename RNG = DefaultRNG>
+    RNGState<RNG> generate_rademacher_vector_r123(sint_t* buff, int64_t n, RNGState<RNG> seed_state) {
```
rileyjmurray, line 31: Actually, when it comes to templating RNGStates, please use the following structure
```
typename state_t = RNGState<DefaultRNG>
```

RandBLAS/trig_skops.hh
```diff
@@ -125,29 +103,29 @@ namespace RandBLAS {
...
             for (int64_t i=0; i < d; i++) {
-                randblas_error_if_msg(selected_rows[i] == top,
-                                      "The list of provided indices should be unique");
                 if (selected_rows[i] != top) {
                     // Use BLAS swap to swap the entire rows
                     // Swapping row 'selected' with row 'top'
                     blas::swap(cols, &A[top], rows, &A[selected_rows[i]], rows);
                 }
+                else
+                    continue;
```
rileyjmurray, line 112: Minor point. Since the "else continue" isn't doing anything functional, please use the following style instead.
```
                if (selected_rows[i] != top) {
                    // Use BLAS swap to swap the entire rows
                    // Swapping row 'selected' with row 'top'
                    blas::swap(cols, &A[top], rows, &A[selected_rows[i]], rows);
                } // else, continue.
```

RandBLAS/trig_skops.hh
```diff
@@ -316,125 +298,157 @@ namespace RandBLAS {
...
-            fht_left_row_major(buff, log_n, num_rows, num_cols);
+            fht_left_row_major(A, log_n, num_rows, num_cols);
         else if(!left && layout == blas::Layout::ColMajor)
-            fht_right_col_major(buff, log_n, num_rows, num_cols);
+            fht_right_col_major(A, log_n, num_rows, num_cols);
         else
-            fht_right_row_major(buff, log_n, num_rows, num_cols);
+            fht_right_row_major(A, log_n, num_rows, num_cols);
     }
 }
```
rileyjmurray, line 318: Please accept num_rows, num_cols, and a leading dimension parameter "lda". (You'll probably need the lda to also appear in the signatures of the functions you're dispatching here.)

RandBLAS/trig_skops.hh
```diff
@@ -316,125 +298,157 @@ namespace RandBLAS {
...
+            diag_scale = new sint_t[m];
+        else
+            diag_scale = new sint_t[n];
+        selected_idxs = new sint_t[d];
+    }
+
+    // Destructor
+    ~HadamardMixingOp() {
+        free(this->diag_scale);
+        free(this->selected_idxs);
```
rileyjmurray, line 350: Use ``delete []``, not ``free``: https://www.geeksforgeeks.org/delete-and-free-in-cpp/.

RandBLAS/trig_skops.hh
```diff
@@ -316,125 +298,157 @@ namespace RandBLAS {
...
+            fht_right_row_major(A, log_n, num_rows, num_cols);
     }
 }
 
 
 namespace RandBLAS::trig {
+template <SignedInteger sint_t = int64_t>
+struct HadamardMixingOp{
+    sint_t* diag_scale;
+    sint_t* selected_idxs;
```
rileyjmurray, line 325: Thinking about it more, I want selected_idxs to always be ``int64_t*``. The reason that sparse matrices use a template parameter ``sint_t`` is because in sparse matrix multiplication moving around indices can actually contribute to the bottleneck runtime. In our context the ``selected_idxs`` is used to specify how we move much larger chunks of data (entire rows in the case of mixing from the left) and so using int64_t doesn't risk making things significantly more expensive.

I'm not sure what to do about the type of ``diag_scale``, to be honest. I won't let this hold back merging the PR but I'd appreciate it if you put thought into this.

RandBLAS/trig_skops.hh
```diff
@@ -316,125 +298,157 @@ namespace RandBLAS {
...
+
+/*
+ * A free-function that performs an inversion of a matrix transformed by `lmiget | rmiget`
+ * the inversion is also performed in-place
+*/
+template <typename T, SignedInteger sint_t = int64_t>
+void invert(
+    HadamardMixingOp<sint_t> &hmo, // details about the transform
+    T* SA // sketched matrix
+) {
```
rileyjmurray, line 364: This type of function name would be problematic if someone wanted to call it from a language that doesn't support function overloading. Please propose some alternative names (here on a comment thread is fine).

RandBLAS/trig_skops.hh
```diff
@@ -316,125 +298,157 @@ namespace RandBLAS {
...
+    int ld = (hmo.left) ? hmo.m : hmo.n;
+
+    T log_sz = std::log2(ld);
+    T log_int_sz, log_final_sz;
+    T log_frac_sz = std::modf(log_sz, &log_int_sz);
+
+    if(log_frac_sz < 1e-3)
+        log_final_sz = log_int_sz;
+    else
+        log_final_sz = log_int_sz + 1;
```
rileyjmurray, line 393: This sort of check makes me very uncomfortable. Please handle the problem of computing $\lceil \log_2(\texttt{ld})\rceil$ in a way that's more robust, or provide a proof that this way of doing this is safe.

RandBLAS/trig_skops.hh
```diff
@@ -316,125 +298,157 @@ namespace RandBLAS {
...
-    sint_t* diag = new sint_t[n];
-    sint_t* selected_rows = new sint_t[d];
-
     auto [ctr, key] = random_state;
 
     //Step 1: Scale with `D`
         //Populating `diag`
-    generate_rademacher_vector_r123(diag, key[0], ctr[0], n);
-    apply_diagonal_rademacher(true, layout, m, n, A, diag);
+    RNGState<RNG> state_idxs = generate_rademacher_vector_r123(hmo.diag_scale, hmo.n, random_state);
```
rileyjmurray, line 419: Use ``auto next_state = generate_rademacher ...``, then reassign ``next_state = repeated_fisher ...``.

test/test_matmul_cores/test_trig.cc
```diff
@@ -27,7 +27,7 @@
 // POSSIBILITY OF SUCH DAMAGE.
 //
 
-#include "RandBLAS.hh"
+// #include "RandBLAS.hh"
```
rileyjmurray, line 30: Don't just comment out. Please delete.

test/test_matmul_cores/test_trig.cc
```diff
@@ -250,41 +250,43 @@ class TestLMIGET : public::testing::Test
         uint32_t seed,
         int64_t m, // Generated data matrix, `A` is of size `(m x n)`
         int64_t n,
-        int64_t d, // #rows/cols that will be permuted
+        int64_t d,
         bool left,
         blas::Layout layout,
-        double epsilon=1e-5
+        T epsilon=RandBLAS::sqrt_epsilon<T>()
```
rileyjmurray, line 256: Rename "epsilon" to "tol"

rileyjmurray (commented):
I left lots of individual comments. After you resolve these, the next step will be to remove dependency on Eigen.

rileyjmurray:
@aryamanjeendgar please sign the Contributor License Agreement.

RandBLAS/trig_skops.hh
aryamanjeendgar: As of right now, I renamed it to `invert_hadamard` --- but maybe since its a free method of the `HadamardMixingOp` class, we can also call it `invert_hmo`?

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,538 @@
...
+    // entries in-place to an input matrix, `A`
+    template<typename T, SignedInteger sint_t = int64_t>
+    void apply_diagonal_rademacher(
+                                bool left, // Pre-multiplying?
+                                blas::Layout layout,
+                                int64_t rows,
+                                int64_t cols,
+                                T* A,
+                                sint_t* diag
+                                ) {
```
rileyjmurray, line 60: fix indentation to be consistent with our conventions. e.g., those here: https://github.com/BallisticLA/RandBLAS/blob/main/RandBLAS/skve.hh

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,538 @@
...
+#include <vector>
+
+#define MAX(a, b) (((a) < (b)) ? (b) : (a))
+#define MIN(a, b) (((a) < (b)) ? (a) : (b))
+
+namespace RandBLAS {
+    // =============================================================================
+    /// WARNING: None of the following functions or overloads thereof are part of the
+    /// public API
+    ///
```
rileyjmurray, line 28: Don't put things in the top-level namespace if they aren't part of the public API. Use a nested namespace. Example: https://github.com/BallisticLA/RandBLAS/blob/fd180c5743757936fdfa5be8b0fecc62d66b55c4/RandBLAS/sparse_data/csc_trsm_impl.hh#L40

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,538 @@
...
+
+    template<typename T, SignedInteger sint_t = int64_t>
+    void permute_rows_to_top(
+                          blas::Layout layout,
+                          int64_t rows,
+                          int64_t cols,
+                          sint_t* selected_rows,
+                          int64_t d, // size of `selected_rows`
+                          T* A
+                          ) {
```
rileyjmurray, line 101: dimensions:
* m, number of rows in A
* n, number of columns in A
* lda, leading dimension of mat(A) when reading from A in "layout" order.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,538 @@
...
+    HadamardMatrixCouple(
+        bool left,
+        blas::Layout layout,
+        int64_t rows,
+        int64_t cols,
+        T* A,
+        int64_t lda,
+        T* workspace
+    ) : left(left), layout(layout), n_rows(rows), n_cols(cols), A(A), lda(lda), workspace(workspace) {};
+};
```
rileyjmurray, line 354: Call this type  ``PaddedMatrix``.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,538 @@
...
+
+/*
+ * A `pure` description of the \PiRHT as a «data-oblvious» transform
+ */
+template <typename T, SignedInteger sint_t = int64_t, typename state_t = RNGState<DefaultRNG>>
+struct HadamardMixingOp{
+    sint_t* diag_scale = nullptr;
+    sint_t* selected_idxs = nullptr;
```
rileyjmurray, line 362: Different name. ``perm_idx``, maybe?

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,538 @@
...
+            hmo.d,
+            hmo.dim,
+            1,
+            hmo.selected_idxs,
+            next_state
+        );
+
+        return next_state;
+    }
+}
```
rileyjmurray, line 429: fill functions that accept a datastructure, like a SketchingOperator, should not return a next state. The next state is computed at construction time with a helper function that performs _no random sampling_. Example:
https://github.com/BallisticLA/RandBLAS/blob/fd180c5743757936fdfa5be8b0fecc62d66b55c4/RandBLAS/sparse_skops.hh#L243-L260

and

https://github.com/BallisticLA/RandBLAS/blob/fd180c5743757936fdfa5be8b0fecc62d66b55c4/RandBLAS/sparse_skops.hh#L376-L384

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,538 @@
...
+/*
+ * Helper function to compute the size of the scratch workspace required
+ * Note, that this follows the same convention as `cuSolver` where the library provides
+ * helper functions for computing the size of scratch workspace required for a driver, often with
+ * a very similar signature to the driver itself
+*/
+template <typename T, typename state_t = RNGState<DefaultRNG>, SignedInteger sint_t = int64_t>
+inline int64_t miget_workspace_sz(
+    HadamardMixingOp<T, sint_t, state_t> &hmo,
+    HadamardMatrixCouple<T> &hmc) {
```
rileyjmurray, line 488: closing paren on new line

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,538 @@
...
+
+/*
+ * Applies an in-place, SRHT-like transform to the input matrix
+ * i.e. A <- (\Pi H D)A OR A <- A(D H \Pi) (which is equivalent to A <- A(\Pi H D)^{-1})
+ * `HadamardMixingOp hmo`:  Data-oblivious description of the \PiRHT at hand
+ * `HadamardMatrixCouple hmc`: Description of the data to be sketched
+ */
+template <typename T, typename state_t = RNGState<DefaultRNG>, SignedInteger sint_t = int64_t>
+inline void miget(
+    HadamardMixingOp<T, sint_t> &hmo, // All information about `A` && the $\mathbb{\Pi\text{RHT}}$
```
rileyjmurray, line 505: ``// All information about `A` && the $\mathbb{\Pi\text{RHT}}$``

^ That's no longer true. Data for ``A`` isn't stored in HMO.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,538 @@
...
+    HadamardMatrixCouple<T> &hmc
+) {
+
+    if(!hmo.filled()) {
+        fill_hadamard<T, state_t>(hmo);
```
rileyjmurray, line 510: Let types be inferred; just ``fill_hadamard(hmo)``.

test/test_matmul_cores/test_trig.cc
```diff
@@ -0,0 +1,645 @@
...
+
+        // Create a vector of the specified size
+        std::vector<T> random_vector(size);
+
+        // Generate random doubles and fill the vector
+        for (int i = 0; i < size; ++i) {
+            random_vector[i] = dist(gen);
+        }
+
+        return random_vector;
```
rileyjmurray, line 115: Don't use STL for random number generation in these contexts. Get random matrices from:

https://github.com/BallisticLA/RandBLAS/blob/main/test/test_matmul_cores/linop_common.hh

test/test_matmul_cores/test_trig.cc
```diff
@@ -0,0 +1,645 @@
...
+                        n,
+                        1.0,
+                        A_vec.data(),
+                        n,
+                        perm_matrix,
+                        n,
+                        0.0,
+                        true_perm_matrix,
+                        n
+                    );
```
rileyjmurray, line 216: Please don't do this .... more than one argument per line is okay here

test/test_matmul_cores/test_trig.cc
```diff
@@ -0,0 +1,645 @@
...
+                    RandBLAS::permute_cols_to_left(layout, m, n, v, 1, A_vec.data());
+                }
+            }
+
+            T norm_permute = 0.0;
+            blas::axpy(m * n, -1.0, A_vec.data(), 1, true_perm_matrix, 1);
+            norm_permute = blas::nrm2(m * n, true_perm_matrix, 1);
+
+            randblas_require(norm_permute < tol);
+
```
rileyjmurray, line 245: Don't check for array equality by checking if norm is (sufficiently) small. Check componentwise so we get more information if/when a test fails. Use
https://github.com/BallisticLA/RandBLAS/blob/fd180c5743757936fdfa5be8b0fecc62d66b55c4/test/comparison.hh#L181-L229

test/test_matmul_cores/test_trig.cc
```diff
@@ -0,0 +1,645 @@
...
+TEST_F(TestLMIGET, test_diag_left_colmajor) {
+    for(uint32_t seed: keys)
+        correctness<double>(
+            seed,
+            transforms::diag_scale,
+            100,
+            100,
+            true,
+            blas::Layout::ColMajor
+        );
```
rileyjmurray, line 455: One line.

rileyjmurray (changes requested):
Woooo progress!

aryamanjeendgar:
@rileyjmurray, the recent commit should address all of your comments in the review!

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,547 @@
...
+state_t generate_rademacher_vector_r123(sint_t* buff, int64_t n, state_t &seed_state) {
+    DefaultRNG rng;
+    auto [ctr, key] = seed_state;
+
+    for (int64_t i = 0; i < n; ++i) {
+        typename DefaultRNG::ctr_type r = rng(ctr, key);
+
+        float rand_value = r123::u01fixedpt<float>(r.v[0]);
+
+        buff[i] = rand_value < 0.5 ? -1 : 1;
```
rileyjmurray, line 40: Conversion to float isn't necessary. Just use ``sint_t rand_value = (r.v[0] % 2 == 0) -1 : 1;``.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,547 @@
...
+///
+
+// Generates a vector of Rademacher entries using the Random123 library
+template<SignedInteger sint_t = int64_t, typename state_t = RNGState<DefaultRNG>>
+state_t generate_rademacher_vector_r123(sint_t* buff, int64_t n, state_t &seed_state) {
+    DefaultRNG rng;
+    auto [ctr, key] = seed_state;
+
+    for (int64_t i = 0; i < n; ++i) {
+        typename DefaultRNG::ctr_type r = rng(ctr, key);
```
rileyjmurray, line 36: This might raise a compiler error if ``state_t`` isn't ``RNGState<DefaultRNG>``.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,547 @@
...
+    }
+    else {
+        // For `RowMajor` ordering
+        for (int64_t i=0; i < d; i++) {
+            if (selected_rows[i] != top) {
+                blas::swap(cols, &A[cols * selected_rows[i]], 1, &A[cols * top], 1);
+            } // else, continue;
+        }
+    }
+}
```
rileyjmurray, line 121: This is missing the leading dimension parameter for ``A``.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,547 @@
...
+else {
+    // For `RowMajor` ordering
+    for (int64_t i=0; i < d; i++) {
+        if (selected_cols[i] != left) {
+            blas::swap(rows, &A[selected_cols[i]], cols, &A[left], cols);
+        }
+    }
+}
+}
```
rileyjmurray, line 151: As above, this is missing ``lda``.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,547 @@
...
+    }
+}
+else {
+    // For `RowMajor` ordering
+    for (int64_t i=0; i < d; i++) {
+        if (selected_cols[i] != left) {
+            blas::swap(rows, &A[selected_cols[i]], cols, &A[left], cols);
+        }
+    }
+}
```
rileyjmurray, line 150: Indent.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,547 @@
...
+template <typename T>
+void fht_dispatch(
+    bool left, // Pre-multiplying?
+    blas::Layout layout,
+    int64_t num_rows,
+    int64_t num_cols,
+    int64_t log_n,
+    T* A,
+    int64_t workspace_ld=0, // leading dimension of the workspace buffer
+    T* workspace_buf=nullptr
```
rileyjmurray, line 307: If ``workspace_buf`` is nullptr then this can segfault. Either make it a required argument or have special logic to handle when it's nullptr.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,547 @@
...
+struct PaddedMatrix {
+    int64_t n_rows;
+    int64_t n_cols;
+    blas::Layout layout;
+    T* data; // Data-matrix
+    int64_t lda;
+    T* work; // Extra space of sz: (2 ** n_closest - lda) * cols/rows
+                  // Memory **has** to be managed by the user
+                  //NOTE: I am also going to assume that the user is kind enough to
+                  // populate this fully with zeros
```
rileyjmurray, line 337: I'd prefer that we not make this assumption. We should zero things out as the first step of applying the RHT.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,547 @@
...
+    PaddedMatrix(
+        bool left,
+        blas::Layout layout,
+        int64_t rows,
+        int64_t cols,
+        T* data,
+        int64_t lda,
+        T* work
+    ) : left(left), layout(layout), n_rows(rows), n_cols(cols), data(data), lda(lda), work(work) {};
+};
```
rileyjmurray, line 352: This needs more comprehensive documentation. Compare to ``SparseSkOp``. Add web docs:
image

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,547 @@
...
+    // Constructor, user-managed memory
+    HadamardMixingOp(int64_t dim,
+                     int64_t d,
+                     sint_t* diag_scale,
+                     sint_t* perm_idx
+                    ) : dim(dim), d(d), diag_scale(diag_scale), perm_idx(perm_idx), seed_state(seed_state), own_memory(false) {};
+
+    bool filled() {
+        if(this -> diag_scale == nullptr \
+        && this -> perm_idx == nullptr)
```
rileyjmurray, line 394: Style: remove ``this -> ``.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,547 @@
...
+    state.counter.incr(d * 1 + dim);
+
+    return state;
+}
+
+/*
+ * A `pure` description of the \PiRHT as a «data-oblvious» transform
+ */
+template <typename T, SignedInteger sint_t = int64_t, typename state_t = RNGState<DefaultRNG>>
+struct HadamardMixingOp {
```
rileyjmurray, line 369: Comprehensive documentation needed.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,547 @@
...
+    auto next_state = RandBLAS::trig::kernels::generate_rademacher_vector_r123(hmo.diag_scale, hmo.dim, hmo.seed_state);
+
+    // Populating `perm_idx`
+    next_state = repeated_fisher_yates<sint_t>(
+        hmo.d,
+        hmo.dim,
+        1,
+        hmo.perm_idx,
+        next_state
+    );
```
rileyjmurray, line 435: one line.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,547 @@
...
+
+    hmo.perm_idx = new sint_t[hmo.d];
+
+    auto [ctr, key] = hmo.seed_state;
+
+    // Populating `diag`
+    auto next_state = RandBLAS::trig::kernels::generate_rademacher_vector_r123(hmo.diag_scale, hmo.dim, hmo.seed_state);
+
+    // Populating `perm_idx`
+    next_state = repeated_fisher_yates<sint_t>(
```
rileyjmurray, line 429: Remove excessive blank lines.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,547 @@
...
+    blas::Layout layout;
+    T* data; // Data-matrix
+    int64_t lda;
+    T* work; // Extra space of sz: (2 ** n_closest - lda) * cols/rows
+                  // Memory **has** to be managed by the user
+                  //NOTE: I am also going to assume that the user is kind enough to
+                  // populate this fully with zeros
+    //NOTE: We could (should?) explore the possibility of managing the buffer by ourselves
+    int64_t work_ld;
+    bool left;
```
rileyjmurray, line 340: I don't like the name ``left`` in this context. I suggest you define an enum (perhaps a static member of this type) that you can use to indicate if you're padding with extra rows or extra columns. As a matter of practice, you'll pad with rows when applying an RHT from the left and you'll pad with columns when applying an RHT from the right.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,547 @@
...
+                     sint_t* perm_idx
+                    ) : dim(dim), d(d), diag_scale(diag_scale), perm_idx(perm_idx), seed_state(seed_state), own_memory(false) {};
+
+    bool filled() {
+        if(this -> diag_scale == nullptr \
+        && this -> perm_idx == nullptr)
+            return false;
+        else
+            return true;
+    }
```
rileyjmurray, line 398: This function can be removed and in-lined with equivalent code
```
bool filled = (hmo.diag_scale != nullptr) && (hmo.perm_idx != nullptr);
```

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,547 @@
...
+    private:
+};
+
+/*
+* Populates a `HadamardMixingOp`: `perm_idx` and `diag_scale`
+*/
+template <typename T, typename state_t = RNGState<DefaultRNG>, SignedInteger sint_t = int64_t>
+void fill_hadamard(
+    HadamardMixingOp<T, sint_t> &hmo
+) {
```
rileyjmurray, line 417: Indent the contents of this function.

RandBLAS/trig_skops.hh
```diff
@@ -0,0 +1,547 @@
...
+
+    //Step 1: Permute the rows/cols
+    if(pm.left)
+        RandBLAS::trig::kernels::permute_rows_to_top(pm.layout, pm.n_rows, pm.n_cols, perm_idx.data(), hmo.d, pm.data);
+    else
+        RandBLAS::trig::kernels::permute_cols_to_left(pm.layout, pm.n_rows, pm.n_cols, perm_idx.data(), hmo.d, pm.data);
+
+    //Step 2: Apply the Hadamard transform (invH = H.T = H)
+    // This has to be a `double` because `blas::scal` really doesn't like being passed integer types
+    double scal_factor = std::pow(2, int(std::log2(hmo.dim)) + 1);
```
rileyjmurray, line 466: Use ``T scal_factor``, not ``double scal_factor``.

rileyjmurray (changes requested):
See individual comments.

rileyjmurray:
@aryamanjeendgar I've reviewed and left more comments.

PR #127, opened 2025-02-15
Adding sparse trsm to the public API

rileyjmurray:
@PTNobel did fantastic work at the AMD Tools workshop on adding sparse trsm kernels to RandBLAS. This PR is to exposes those kernels to RandBLAS' public API.

Changes
 * Add ``const`` qualifiers in sparse matrix format conversion functions.
 * Add a utility function, ``compressed_indices_are_increasing``, to sparse_data/base.hh for use in checking assumptions in the new sparse trsm function.
 * Fix minor documentation mistakes (arising from copy-pasting ...) in sparse_data/csc_matrix.hh.
 * Add ``diag`` flags to sparse trsm kernels and ``nonunit`` flags to sparse trsv kernels. Remove redundant dimensional arguments from these kernels. In TRSM, change the variable used to refer to the number of right-hand-sides from "k" to "n" (and change the square matrix dimension from "n" to "m").
 * Updated sparse TRSM tests to use the new public-API sparse_data::trsm function rather than private implementation functions. Add tests for the transposition flag in sparse_data::trsm.

### DIFF
--- a/RandBLAS/sparse_data/base.hh
+++ b/RandBLAS/sparse_data/base.hh
@@ -32,6 +32,7 @@
 #include "RandBLAS/base.hh"
 #include <blas.hh>
 #include <concepts>
+#include <iostream>
 
 
 namespace RandBLAS::sparse_data {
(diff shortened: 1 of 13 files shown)


rileyjmurray:
Plan:
* Stick with the current implementation that requires nonzero indices for a row in CSR/column in CSC to be sorted. Have an optional argument indicating if we should check this condition on entry, with a default to perform the check. (The optional argument might not be a boolean so I won't say "a default of true.")
* Open a GitHub issue about adding kernels that relax the above requirement and behave more like MKL/AOCL.

PR #133, opened 2025-08-13
Concept Pre-Processors 

OG-ttafolla:
added pre-processor commands for concepts in order to run with older compilers


rileyjmurray (approved):
This looks great, @OG-ttafolla! Future users of RandBLAS owe you a debt of saving them compiler pain.

(Also, this has taught me how to properly use preprocessor directives that affect function templating.)

PR #137, opened 2025-10-10
Improvements to API for sparse matrices

### LINKED ISSUE #125: reserve_coo should zero-initialize all memory
This is a minor issue. Debugging would be easier if we zero-initialized memory here:
https://github.com/BallisticLA/RandBLAS/blob/dea0ae09e84ea20a1ffec757cf09e33b32ac1ab3/RandBLAS/sparse_data/coo_matrix.hh#L285-L287

### LINKED ISSUE #131: Sparse sketching on row-major data dispatches inefficient kernel
We implement sketching with a SparseSkOp by creating a COOMatrix view and then dispatching left_spmm or right_spmm, where a call to the latter reduces to a call to the former.

The implementation of left_spmm only has one code path for COOMatrix objects: apply_coo_left_jki_p11. This applies a bespoke COO-CSC conversion that lets us select a submatrix of the COOMatrix. After that, we call apply_csc_left_jki_p11, which parallelizes over columns. The last of these steps is extremely inefficient for row-major data; we should be dispatching apply_csc_left_kib_rowmajor_1p1.

This is somewhat related to #128.

rileyjmurray:
Everything in the plan below is implemented and tested. 

#### Plan

 * [x] make sparse matrix arguments const unless mutation is clearly necessary.
    * [x] sketching functions
    * [x] spmm dispatch functions
    * [x] coo spmm kernel (rather, dispatcher to CSC kernels).
    * [x] coo_to_cs[r/c] conversion functions.
 * [x]  dispatch efficient matrix-multiply for sparse sketching with row-major data (will resolve #131).
 * [x] Allocation of COO sparse matrix data should zero-initialize all arrays. (resolves #125)
 * [x] utility function for applying a symmetric permutation to a COOMatrix.
 * [x] add ``deepcopy()`` instance methods and free functions.
     * [x] CSCMatrix
     * [x] CSRMatrix
     * [x] COOMatrix
 * [x]  transpose instance methods for sparse matrix classes.
    * [x] CSCMatrix
    * [x] CSRMatrix
    * [x] COOMatrix
  * [x] eliminate allocation of temporary ``vals`` arrays in spmm kernels when ``alpha != 0``. 
 * [x] format conversion instance methods
    * [x] COO --> CSC or CSR. 
    * [x] CSR --> COO
    * [x] CSC --> COO
  * [x] COOMatrix instance method for sorting 

#### other changes

 * SparseDist and DenseDist
    * have constructors that accept any ordinals for dimensional arguments, rather than only int64_t. We round down floats as needed.  
    * add ``sample(RNGState &seed_state)`` instance functions.

### DIFF
--- a/rtd/source/tutorial/sketch_updates.rst
+++ b/rtd/source/tutorial/sketch_updates.rst
@@ -206,9 +206,9 @@ Implementation
       SkDist D(  d, m + k, arg3, Axis::Short ); 
       // Since d < m and we're short-axis major, the columns of matrices sampled from
       // D1 or D1 will be sampled i.i.d. from some distribution on R^d.
-      using SkOp = typename SkDist::distribution_t;
-      SkOp S1( D1,  seed_state ); // seed_state is some RNGState.
-      SkOp  S(  D,  seed_state );
+
+      auto S1 = D1.sample( seed_state ); // seed_state is some RNGState.
+      auto S  =  D.sample( seed_state );
       // With these definitions, S1 is *always* equal to the first m columns of S.
       // We recover S2 by working implicitly with the trailing k columns of S.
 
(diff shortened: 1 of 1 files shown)


RandBLAS/sparse_data/sksp.hh
rileyjmurray: Docstrings were updated to ``const SpMat`` but the actual function signatures were not.

PR #163, opened 2026-05-13
Add SYMM kernels for sparse symmetric matrices; rework sketch_symmetric to exploit symmetry

mmelnich:
## Summary

Adds **SYMM-shaped kernels** to RandBLAS so that symmetric matrices can finally be sketched with their symmetry exploited. Implements all four cases (A, B, C, D), and folds in repo-wide cleanups uncovered along the way (a `lascl` utility, helper extractions, a CSC fast path).

Drives the funNyström++ work on `BallisticLA/RandLAPACK#132`: `sketch_symmetric` was silently forwarding to `sketch_general` → GEMM, so A's symmetry wasn't exploited and a sparse SkOp had to be densified before the call. This PR closes the dense side and lays the foundation for the sparse-symmetric side.

Design plan: [`randblas-symm-plan.md`](https://github.com/BallisticLA/RandBLAS/blob/add-symm-kernels/Claude_files/randnla/project-plans/randblas-symm-plan.md). Condensed 4-case table + MKL availability matrix lives in [`RandBLAS/sparse_data/DevNotes.md`](RandBLAS/sparse_data/DevNotes.md).

## The four cases

| Tag | Operation | A storage | B storage | Status |
|-----|-----------|-----------|-----------|--------|
| **A** | dense-symm × dense | dense, one triangle | dense | **Implemented** via `blas::symm` in `sksy.hh` |
| **B** | dense-symm × sparse | dense, one triangle | sparse | **Implemented**. `lsksys` / `rsksys` wrappers in `sksy.hh` handle the SparseSkOp materialization-and-recurse; the COO walk plus two-axpy-per-stored-nonzero scatter lives in `RandBLAS::sparse_data::coo_lsksys` / `coo_rsksys` in `sparse_data/coo_sksys_impl.hh` and reads only A's named triangle |
| **C** | sparse-symm × dense → dense | sparse, one triangle | dense | **Implemented** in `sparse_data/spsymm_dispatch.hh` (MKL fast path covers all of `{CSR, CSC, COO} × {side=Left, side=Right}`; per-format hand kernels remain as the non-MKL fallback) |
| **D** | sparse-symm × sparse → dense | sparse, one triangle | sparse | **Implemented** via densify-B + Case-C composition in `sparse_data/spsymm_dispatch.hh` (new `spsymm` overload taking two `SparseMatrix` args) |

## What's in the PR

**Cases A, B, C, all real implementations**. `sketch_symmetric` dispatches by SkOp type. DenseSkOp goes through `dense::lsksy3`/`rsksy3` (SYMM-backed, with transpose-copy into matching layout on layout-mismatch so the SYMM speedup is retained). SparseSkOp goes through `sparse::lsksys`/`rsksys` wrappers in `sksy.hh`, which handle the SparseSkOp materialization-and-recurse and forward into `RandBLAS::sparse_data::coo_lsksys`/`coo_rsksys` in `sparse_data/coo_sksys_impl.hh`; the kernel does a per-stored-nonzero two-AXPY scatter that reads only A's named triangle, with submatrix offsets filtered inline on the COO view. `spsymm` dispatches to an MKL fast path covering both Side values: side=Left direct, side=Right via opposite-layout reinterpretation of B/Y (no-copy trick exploiting `A == A^T`), and CSC via `A.transpose()` lightweight CSR view plus flipped `uplo`. The hand-rolled per-format fallback (`csr/csc/coo_spsymm_impl.hh`) remains as the non-MKL path; its per-stored-entry scatter is shared via `internal::spsymm_scatter_{left,right}` helpers.

**Case D, densify-B plus Case-C composition**. The new two-`SparseMatrix`-arg `spsymm` overload allocates an `m`-by-`n` temporary `std::vector<T>` in the caller's layout, fills it via the format-specific `coo_to_dense` / `csr_to_dense` / `csc_to_dense` helper picked by `if constexpr`, and then calls the existing Case-C dispatcher on the densified buffer. Covers all 3 × 3 = 9 sparse-format pairings for `(A, B)`. MKL's `mkl_sparse_sp2m` returns `SPARSE_STATUS_NOT_SUPPORTED` when `descrA.type == SPARSE_MATRIX_TYPE_SYMMETRIC` (only `GENERAL` is accepted there), and `mkl_sparse_d_spmmd` takes no descriptor at all, so the symmetric expansion has to happen on the RandBLAS side either way. Cost: `O(m * n)` temporary buffer, small for the typical RandNLA workload where `B` is a sketching operator with `nnz(B) << m*n`.

**`Symmetric<SpMat>` wrapper** (`sparse_data/symmetric.hh`): non-owning carrier holding `const SpMat&` and `blas::Uplo`. Square-matrix invariant at construction. Type-safety guard: a `Symmetric<SpMat>` argument fails to bind to general `spmm`/`spgemm`. Public `RandBLAS::spsymm` has both a raw (`SpMat`+`uplo`) overload and a wrapper-taking overload.

**Cleanups along the way**:

- **`RandBLAS::util::lascl`** (`util.hh`): `A := α·A` for a dense m×n matrix, LAPACK-style naming. Replaces hand-rolled "loop-over-`blas::scal`" beta-scale patterns at 7 sites across `trsm_dispatch.hh`, `spmm_dispatch.hh`, `mkl_spmm_impl.hh` (`mkl_spgemm_to_dense`), the three new `*_spsymm_impl.hh` files, and two example programs.
- **`internal::mkl_sparse_mm_call`** (`mkl_spmm_impl.hh`): type-dispatched wrapper around `mkl_sparse_d_mm`/`mkl_sparse_s_mm`, shared by `mkl_left_spmm` (GENERAL descriptor) and `mkl_spsymm` (SYMMETRIC descriptor).
- **`<numeric>` include** in `sparse_data/base.hh`: latent gap (used `std::iota` but had relied on transitive inclusion via `coo_matrix.hh`); the new `symmetric.hh` exposed it.
- **`RandBLAS.hh` umbrella**: now pulls in `sparse_data/spsymm_dispatch.hh` so downstream `#include <RandBLAS.hh>` consumers see `RandBLAS::spsymm` and the `Symmetric<SpMat>` wrapper automatically (mirrors how `spmm`/`spgemm`/`sketch_symmetric` are already exposed).

**Perf benchmark**: new `examples/simple-kernel-benchmarks/spsymm_performance.cc` compares (a) sparse: `RandBLAS::spsymm` (one-triangle storage + MKL fast path) vs the pre-PR workaround (both triangles + `RandBLAS::spmm`) vs a dense `blas::symm` reference, and (b) dense: the new `sketch_symmetric` (SYMM-backed) vs the equivalent `sketch_general` call (the pre-PR GEMM-forwarding behaviour). Single-config CLI or default 3-point sweep.

## API break worth flagging

**`sketch_symmetric` now takes `blas::Uplo uplo`** (previously: `T sym_check_tol = 0`). All four overloads were affected. Justification:

- Pre-PR, `sketch_symmetric` performed an O(n²) runtime symmetry check via `require_symmetric`, then forwarded to GEMM. The check was a workaround for the fact that GEMM reads both triangles.
- With SYMM dispatch via `blas::symm`, only the named triangle is read. The check is meaningless and pure overhead.
- `uplo` is the BLAS-standard way to name which triangle is read.

In-tree callers patched in this PR: `test/linops/test_sketch_symmetric.cc`. The only known external consumer is RandLAPACK's `ExplicitSymLinOp` (via the funNyström++ branch); follow-up RandLAPACK PR will thread `uplo` through. `require_symmetric` remains available as a standalone validator.

## What's still missing

Deliberate scope cuts; not blockers for this PR.

- **Hand-rolled Case D kernel** (e.g., a true sparse-symm × sparse → dense scatter that walks A's stored triangle and scatters into Y via a sparse-RHS gather): could in principle skip the dense temp buffer for `B`, but the dispatch grid is messier (mixed A/B sparse formats × side × layout × uplo) and no portable reference exists in MKL / Ginkgo / SparseBLAS reference / MAGMA-sparse (surveyed 2026-05). The densify-B + Case-C composition shipped here is correct, covers all 9 format pairings, and has a small constant-factor overhead in the regime that matters for RandNLA workloads. If the dense-temp cost becomes a bottleneck, a hand-rolled kernel can be a targeted follow-up.
- **RandLAPACK companion patch**: `ExplicitSymLinOp` and any other `sketch_symmetric` consumers need `uplo` threaded through. Small follow-up; will open after this PR merges.

## Test coverage

24-cell `{COO, CSR, CSC} × {ColMajor, RowMajor} × {Upper, Lower} × {Left, Right}` for `spsymm` Case C (in `test/linops/test_spsymm.cc`), plus float/`beta=0`/`alpha=0`/wrapper-routing edge cases, plus 7 Case D tests (`{CSR-CSR, CSC-CSC, COO-COO, Mixed-format} × Left`, plus `CSR-CSR-Right`, `Float`, and `AlphaZero-BetaScale`), plus 5 cases for the `Symmetric<SpMat>` wrapper, plus 14 cases in `test_sketch_symmetric.cc` (8 DenseSkOp paths + 6 new SparseSkOp paths exercising Case B across both sides, both layouts, both uplo, and a lift configuration).

Tolerance for `spsymm` tests: `atol = 100*eps`, `rtol = 10*eps`. The dense reference (`blas::symm` on a fully-symmetrized A) and the sparse path accumulate FMAs in a different order (off-diagonal entries contribute via two AXPYs in the sparse path vs. one fused dotprod in dense SYMM), giving a few-ULP divergence. The relaxed tolerance documents this is expected, not a bug.

Verified locally: **477/477 ctest pass** on Linux + GCC 13.3 + CUDA-aware blaspp + MKL sparse (includes 17/17 TestSpsymm = 10 Case C + 7 Case D, plus 14/14 TestSketchSymmetric = 8 DenseSkOp + 6 SparseSkOp).

### DIFF
--- a/RandBLAS/sksy.hh
+++ b/RandBLAS/sksy.hh
@@ -38,7 +38,8 @@
 // Symmetric sketching helpers (SYMM-backed). See project-plans/randblas-symm-plan.md
 // for the four-case API design and the implementation status of each case.
 //   - lsksy3, rsksy3: Case A (dense-symm A x dense Omega), via blas::symm.
-//   - SparseSkOp branches of sketch_symmetric: Case B stubs (not implemented).
+//   - lsksys, rsksys: Case B (dense-symm A x sparse SkOp), per-stored-entry
+//     two-axpy scatter that reads only the uplo triangle of A.
 // =============================================================================
 
 namespace RandBLAS::dense {
@@ -187,33 +188,171 @@ void rsksy3(
 } // end namespace RandBLAS::dense
 
 
+namespace RandBLAS::sparse {
+
+// =============================================================================
+/// LSKSYS: dense symmetric A on the right of a SparseSkOp.
+///
+/// Computes B = alpha * submat(S) * mat(A) + beta * B, where:
+///   - mat(A) is n-by-n dense symmetric, with only the `uplo` triangle stored.
+///   - submat(S) is the d-by-n view of S at (ro_s, co_s); S is a SparseSkOp.
+///   - mat(B) is d-by-n dense.
+///
+/// Inner loop: for each stored nonzero (row_S, col_S, v) of S inside the
+/// submatrix window, contribute alpha*v * (row col_S of A) to row (row_S - ro_s)
+/// of B. The row of symmetric A is assembled from the stored triangle as two
+/// blas::axpy calls (one along the stored column, one along the stored row past
+/// the diagonal).
+template <typename T, typename RNG, SignedInteger sint_t>
+void lsksys(
+    blas::Layout layout,
+    blas::Uplo uplo,
+    int64_t d,
+    int64_t n,
+    T alpha,
+    const SparseSkOp<T, RNG, sint_t> &S,
+    int64_t ro_s,
+    int64_t co_s,
+    const T *A,
+    int64_t lda,
+    T beta,
+    T *B,
+    int64_t ldb
+) {
+    if (S.nnz < 0) {
+        SparseSkOp<T, RNG, sint_t> shallowcopy(S.dist, S.seed_state);
+        fill_sparse(shallowcopy);
+        lsksys(layout, uplo, d, n, alpha, shallowcopy, ro_s, co_s, A, lda, beta, B, ldb);
+        return;
+    }
+
+    util::lascl(layout, d, n, beta, B, ldb);
+    if (alpha == T(0)) return;
+
+    auto Scoo = coo_view_of_skop(S);
+    const bool col_major = (layout == blas::Layout::ColMajor);
+
+    for (int64_t p = 0; p < Scoo.nnz; ++p) {
+        sint_t row_S = Scoo.rows[p];
+        sint_t col_S = Scoo.cols[p];
+        if (row_S < ro_s || row_S >= ro_s + d) continue;
+        if (col_S < co_s || col_S >= co_s + n) continue;
+        int64_t i = static_cast<int64_t>(row_S) - ro_s;  // B row
+        int64_t j = static_cast<int64_t>(col_S) - co_s;  // A row index (sym A read as row j)
+        T av = alpha * Scoo.vals[p];
+
+        // B[i, :] += av * row_j(A_sym), split by uplo into two contiguous
+        // ranges of A so each becomes a single axpy.
+        if (uplo == blas::Uplo::Upper) {
+            // row j of A:
+            //   c in [0, j]: read A[c, j]      (column j of stored Upper, rows 0..j)
+            //   c in (j, n-1]: read A[j, c]    (row j of stored Upper, cols j+1..n-1)
+            if (col_major) {
+                blas::axpy(j + 1, av, &A[j * lda], 1, &B[i], ldb);
+                blas::axpy(n - j - 1, av, &A[j + (j + 1) * lda], lda, &B[i + (j + 1) * ldb], ldb);
+            } else {
+                blas::axpy(j + 1, av, &A[j], lda, &B[i * ldb], 1);
+                blas::axpy(n - j - 1, av, &A[j * lda + j + 1], 1, &B[i * ldb + j + 1], 1);
+            }
+        } else {
+            // Lower-stored:
+            //   c in [0, j]: read A[j, c]      (row j of stored Lower, cols 0..j)
+            //   c in (j, n-1]: read A[c, j]    (column j of stored Lower, rows j+1..n-1)
+            if (col_major) {
+                blas::axpy(j + 1, av, &A[j], lda, &B[i], ldb);
+                blas::axpy(n - j - 1, av, &A[(j + 1) + j * lda], 1, &B[i + (j + 1) * ldb], ldb);
+            } else {
+                blas::axpy(j + 1, av, &A[j * lda], 1, &B[i * ldb], 1);
+                blas::axpy(n - j - 1, av, &A[(j + 1) * lda + j], lda, &B[i * ldb + j + 1], 1);
+            }
+        }
+    }
+}
+
+
+// =============================================================================
+/// RSKSYS: dense symmetric A on the left of a SparseSkOp.
+///
+/// Computes B = alpha * mat(A) * submat(S) + beta * B, where:
+///   - mat(A) is n-by-n dense symmetric, with only the `uplo` triangle stored.
+///   - submat(S) is the n-by-d view of S at (ro_s, co_s); S is a SparseSkOp.
+///   - mat(B) is n-by-d dense.
+///
+/// Same two-axpy scatter pattern as lsksys, but on columns of A (the column
+/// of A indexed by row_S of S contributes into the column of B indexed by
+/// col_S - co_s).
+template <typename T, typename RNG, SignedInteger sint_t>
+void rsksys(
+    blas::Layout layout,
+    blas::Uplo uplo,
+    int64_t n,
+    int64_t d,
+    T alpha,
+    const T *A,
+    int64_t lda,
+    const SparseSkOp<T, RNG, sint_t> &S,
+    int64_t ro_s,
+    int64_t co_s,
+    T beta,
+    T *B,
+    int64_t ldb
+) {
+    if (S.nnz < 0) {
+        SparseSkOp<T, RNG, sint_t> shallowcopy(S.dist, S.seed_state);
+        fill_sparse(shallowcopy);
+        rsksys(layout, uplo, n, d, alpha, A, lda, shallowcopy, ro_s, co_s, beta, B, ldb);
+        return;
+    }
+
+    util::lascl(layout, n, d, beta, B, ldb);
+    if (alpha == T(0)) return;
+
+    auto Scoo = coo_view_of_skop(S);
+    const bool col_major = (layout == blas::Layout::ColMajor);
+
+    for (int64_t p = 0; p < Scoo.nnz; ++p) {
+        sint_t row_S = Scoo.rows[p];
+        sint_t col_S = Scoo.cols[p];
+        if (row_S < ro_s || row_S >= ro_s + n) continue;
+        if (col_S < co_s || col_S >= co_s + d) continue;
+        int64_t i = static_cast<int64_t>(row_S) - ro_s;  // A column index
+        int64_t j = static_cast<int64_t>(col_S) - co_s;  // B column
+        T av = alpha * Scoo.vals[p];
+
+        // B[:, j] += av * col_i(A_sym), split by uplo:
+        if (uplo == blas::Uplo::Upper) {
+            // col i of A:
+            //   r in [0, i]: read A[r, i]    (column i of stored Upper, rows 0..i)
+            //   r in (i, n-1]: read A[i, r]  (row i of stored Upper, cols i+1..n-1)
+            if (col_major) {
+                blas::axpy(i + 1, av, &A[i * lda], 1, &B[j * ldb], 1);
+                blas::axpy(n - i - 1, av, &A[i + (i + 1) * lda], lda, &B[(i + 1) + j * ldb], 1);
+            } else {
+                blas::axpy(i + 1, av, &A[i], lda, &B[j], ldb);
+                blas::axpy(n - i - 1, av, &A[i * lda + i + 1], 1, &B[(i + 1) * ldb + j], ldb);
+            }
+        } else {
+            // Lower-stored col i:
+            //   r in [0, i-1]: read A[i, r]  (row i of stored Lower, cols 0..i-1)
+            //   r in [i, n-1]: read A[r, i]  (column i of stored Lower, rows i..n-1)
+            if (col_major) {
+                blas::axpy(i, av, &A[i], lda, &B[j * ldb], 1);
+                blas::axpy(n - i, av, &A[i + i * lda], 1, &B[i + j * ldb], 1);
+            } else {
+                blas::axpy(i, av, &A[i * lda], 1, &B[j], ldb);
+                blas::axpy(n - i, av, &A[i * lda + i], lda, &B[i * ldb + j], ldb);
+            }
+        }
+    }
+}
+
+} // end namespace RandBLAS::sparse
+
+
 namespace RandBLAS {
 
 using namespace RandBLAS::dense;
 using namespace RandBLAS::sparse;
-
-
-namespace detail {
-
-// =============================================================================
-/// Shared Case-B stub-throw helper for the four SparseSkOp-taking
-/// sketch_symmetric specializations. The variadic parameter pack consumes
-/// the caller's arguments so the compiler doesn't warn about unused
-/// parameters in the (genuinely-unused, by design) stub bodies.
-template <typename ...Args>
-[[noreturn]] inline void throw_sketch_symmetric_case_b(Args&&...) {
-    randblas_require(
-        false &&
-        "RandBLAS::sketch_symmetric with a sparse sketching operator is the "
-        "Case-B kernel from project-plans/randblas-symm-plan.md. Not implemented "
-        "in this PR --- the API signature is reserved so future PRs can fill in "
-        "the body without breaking source compatibility. Composition fallback: "
-        "densify the SkOp then call sketch_symmetric on the dense version."
-    );
-    __builtin_unreachable();
-}
-
-} // namespace detail
 
 
 // MARK: SUBMAT(S)
@@ -343,7 +482,7 @@ inline void sketch_symmetric(
     T beta,
     T* B, int64_t ldb
 ) {
-    detail::throw_sketch_symmetric_case_b(layout, uplo, d, n, alpha, S, ro_s, co_s, A, lda, beta, B, ldb);
+    RandBLAS::sparse::lsksys(layout, uplo, d, n, alpha, S, ro_s, co_s, A, lda, beta, B, ldb);
 }
 
 
@@ -402,7 +541,7 @@ inline void sketch_symmetric(
     T beta,
     T* B, int64_t ldb
 ) {
-    detail::throw_sketch_symmetric_case_b(layout, uplo, n, d, alpha, A, lda, S, ro_s, co_s, beta, B, ldb);
+    RandBLAS::sparse::rsksys(layout, uplo, n, d, alpha, A, lda, S, ro_s, co_s, beta, B, ldb);
 }
 
 
@@ -460,7 +599,9 @@ inline void sketch_symmetric(
     T beta,
     T* B, int64_t ldb
 ) {
-    detail::throw_sketch_symmetric_case_b(layout, uplo, alpha, S, A, lda, beta, B, ldb);
+    int64_t d = S.dist.n_rows;
+    int64_t n = S.dist.n_cols;
+    RandBLAS::sparse::lsksys(layout, uplo, d, n, alpha, S, 0, 0, A, lda, beta, B, ldb);
 }
 
 
@@ -515,7 +656,9 @@ inline void sketch_symmetric(
     T beta,
     T* B, int64_t ldb
 ) {
-    detail::throw_sketch_symmetric_case_b(layout, uplo, alpha, S, A, lda, beta, B, ldb);
+    int64_t n = S.dist.n_rows;
+    int64_t d = S.dist.n_cols;
+    RandBLAS::sparse::rsksys(layout, uplo, n, d, alpha, A, lda, S, 0, 0, beta, B, ldb);
 }
 
 } // end namespace RandBLAS
--- a/RandBLAS/sksy.hh
+++ b/RandBLAS/sksy.hh
@@ -180,6 +180,29 @@ using namespace RandBLAS::dense;
 using namespace RandBLAS::sparse;
 
 
+namespace detail {
+
+// =============================================================================
+/// Shared Case-B stub-throw helper for the four SparseSkOp-taking
+/// sketch_symmetric specializations. The variadic parameter pack consumes
+/// the caller's arguments so the compiler doesn't warn about unused
+/// parameters in the (genuinely-unused, by design) stub bodies.
+template <typename ...Args>
+[[noreturn]] inline void throw_sketch_symmetric_case_b(Args&&...) {
+    randblas_require(
+        false &&
+        "RandBLAS::sketch_symmetric with a sparse sketching operator is the "
+        "Case-B kernel from project-plans/randblas-symm-plan.md. Not implemented "
+        "in this PR --- the API signature is reserved so future PRs can fill in "
+        "the body without breaking source compatibility. Composition fallback: "
+        "densify the SkOp then call sketch_symmetric on the dense version."
+    );
+    __builtin_unreachable();
+}
+
+} // namespace detail
+
+
 // MARK: SUBMAT(S)
 
 // =============================================================================
@@ -307,17 +330,7 @@ inline void sketch_symmetric(
     T beta,
     T* B, int64_t ldb
 ) {
-    (void) layout; (void) uplo; (void) d; (void) n; (void) alpha;
-    (void) S; (void) ro_s; (void) co_s; (void) A; (void) lda;
-    (void) beta; (void) B; (void) ldb;
-    randblas_require(
-        false &&
-        "RandBLAS::sketch_symmetric with a sparse sketching operator is the "
-        "Case-B kernel from project-plans/randblas-symm-plan.md. Not implemented "
-        "in this PR --- the API signature is reserved so future PRs can fill in "
-        "the body without breaking source compatibility. Composition fallback: "
-        "densify the SkOp then call sketch_symmetric on the dense version."
-    );
+    detail::throw_sketch_symmetric_case_b(layout, uplo, d, n, alpha, S, ro_s, co_s, A, lda, beta, B, ldb);
 }
 
 
@@ -376,14 +389,7 @@ inline void sketch_symmetric(
     T beta,
     T* B, int64_t ldb
 ) {
-    (void) layout; (void) uplo; (void) n; (void) d; (void) alpha;
-    (void) S; (void) ro_s; (void) co_s; (void) A; (void) lda;
-    (void) beta; (void) B; (void) ldb;
-    randblas_require(
-        false &&
-        "RandBLAS::sketch_symmetric with a sparse sketching operator is the "
-        "Case-B kernel from project-plans/randblas-symm-plan.md. Not implemented."
-    );
+    detail::throw_sketch_symmetric_case_b(layout, uplo, n, d, alpha, A, lda, S, ro_s, co_s, beta, B, ldb);
 }
 
 
@@ -441,14 +447,7 @@ inline void sketch_symmetric(
     T beta,
     T* B, int64_t ldb
 ) {
-    (void) layout; (void) uplo; (void) alpha;
-    (void) S; (void) A; (void) lda;
-    (void) beta; (void) B; (void) ldb;
-    randblas_require(
-        false &&
-        "RandBLAS::sketch_symmetric with a sparse sketching operator is the "
-        "Case-B kernel from project-plans/randblas-symm-plan.md. Not implemented."
-    );
+    detail::throw_sketch_symmetric_case_b(layout, uplo, alpha, S, A, lda, beta, B, ldb);
 }
 
 
@@ -503,14 +502,7 @@ inline void sketch_symmetric(
     T beta,
     T* B, int64_t ldb
 ) {
-    (void) layout; (void) uplo; (void) alpha;
-    (void) S; (void) A; (void) lda;
-    (void) beta; (void) B; (void) ldb;
-    randblas_require(
-        false &&
-        "RandBLAS::sketch_symmetric with a sparse sketching operator is the "
-        "Case-B kernel from project-plans/randblas-symm-plan.md. Not implemented."
-    );
+    detail::throw_sketch_symmetric_case_b(layout, uplo, alpha, S, A, lda, beta, B, ldb);
 }
 
 } // end namespace RandBLAS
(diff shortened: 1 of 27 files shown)


mmelnich:
@rileyjmurray, this is still a WIP, but feel free to look.

RandBLAS/sparse_data/csr_spsymm_impl.hh
```diff
@@ -0,0 +1,151 @@
...
+        return;
+    }
+    if (layout == blas::Layout::ColMajor) {
+        for (int64_t j = 0; j < n; ++j)
+            blas::scal(m, beta, Y + j*ldy, 1);
+    } else {
+        for (int64_t i = 0; i < m; ++i)
+            blas::scal(n, beta, Y + i*ldy, 1);
+    }
+}
```
rileyjmurray: Rename to `lascal`, to follow the LAPACK convention. Move to randblas/utilities. Update functions throughout RandBLAS to use this function rather than whatever hand-rolled thing they currently use.

rileyjmurray:
Regarding LASCAL, there are a few other sites where it can be used that weren't in your most recent commit. You can find some hits in the search

https://github.com/search?q=repo%3ABallisticLA%2FRandBLAS++RandBLAS%3A%3Autil%3A%3Asafe_scal&type=code.

See also https://github.com/BallisticLA/RandBLAS/blob/d8a153fcb94b204fb21cc26b7fb39450a915c143/RandBLAS/sparse_data/mkl_spmm_impl.hh#L379-L383.

RandBLAS/sparse_data/csr_spsymm_impl.hh
mmelnich: Done

mmelnich:
@rileyjmurray lmk your thoughts

rtd/source/api_reference/sketch_sparse.rst
```diff
@@ -114,6 +114,47 @@ Deterministic operations
...
+    for the fast path; otherwise it uses a hand-rolled per-format kernel
+    (CSR / CSC / COO). See ``RandBLAS/sparse_data/DevNotes.md`` for the dispatch
+    flow and the broader 4-case design that ``spsymm`` belongs to.
+
+    .. note::
+
+       The ``Symmetric<SpMat>`` carrier (a non-owning wrapper holding a
+       ``const SpMat &`` and a ``blas::Uplo``) is the type-safety hook that
+       prevents a symmetric sparse matrix from being accidentally passed to
+       the general ``spmm`` / ``spgemm`` routines.
```
rileyjmurray: the "symmetric sparse matrix" mentioned here means "a matrix that the user interprets as symmetric while only storing one of the two triangles", right?

RandBLAS/sparse_data/DevNotes.md
```diff
@@ -54,6 +54,110 @@ either ``lskges`` or ``rskges``. Here's what would happen after we entered one o
...
+RandBLAS exposes a SYMM-style API for sparse symmetric matrices via the
+``spsymm`` family. The design covers four cases based on the structure of the
+two operands (``A`` symmetric vs. the second factor ``B``):
+
+| Tag | Operation                       | A storage           | B storage | Status this PR                                  |
+|-----|---------------------------------|---------------------|-----------|-------------------------------------------------|
+| A   | dense-symm × dense              | dense, one triangle | dense     | Implemented via ``blas::symm`` in ``sksy.hh``.   |
+| B   | dense-symm × sparse             | dense, one triangle | sparse    | Implemented via hand-rolled ``lsksys`` / ``rsksys`` in ``sksy.hh`` (two-axpy scatter per stored nonzero of S, reading only the named triangle of A). |
+| C   | sparse-symm × dense (→ dense)   | sparse, one triangle | dense     | Implemented in ``spsymm_dispatch.hh`` (MKL fast path + per-format fallbacks). |
+| D   | sparse-symm × sparse → dense    | sparse, one triangle | sparse    | Implemented via densify-B + Case-C composition in ``spsymm_dispatch.hh``. |
```
rileyjmurray: This would be the only place where RandBLAS supports sparse-times-sparse. That kind of feature needs to be introduced in a separate PR.

RandBLAS/sksy.hh
```diff
@@ -33,238 +33,398 @@
...
+    T* S_ptr = &S.buff[pos];
+
+    if (S.layout == layout) {
+        // Fast path: SYMM directly.
+        blas::symm(layout, blas::Side::Right, uplo, d, n,
+                   alpha, A, lda, S_ptr, lds, beta, B, ldb);
+    } else {
+        // Layout mismatch: transpose-copy S into a tight buffer in the
+        // caller's layout, then SYMM. Costs O(d*n) for the copy, but keeps
+        // the SYMM speedup (~1.3-1.8x over GEMM) on the d*n*n_A matvec.
```
rileyjmurray: Where does this multiplier come from?

RandBLAS/sksy.hh
```diff
@@ -33,238 +33,398 @@
...
+            //   c in [0, j]: read A[j, c]      (row j of stored Lower, cols 0..j)
+            //   c in (j, n-1]: read A[c, j]    (column j of stored Lower, rows j+1..n-1)
+            if (col_major) {
+                blas::axpy(j + 1, av, &A[j], lda, &B[i], ldb);
+                blas::axpy(n - j - 1, av, &A[(j + 1) + j * lda], 1, &B[i + (j + 1) * ldb], ldb);
+            } else {
+                blas::axpy(j + 1, av, &A[j * lda], 1, &B[i * ldb], 1);
+                blas::axpy(n - j - 1, av, &A[(j + 1) * lda + j], lda, &B[i * ldb + j + 1], 1);
+            }
+        }
```
rileyjmurray: This is a COOMatrix kernel. It should live somewhere in RandBLAS/sparse_data.

RandBLAS/sksy.hh
```diff
@@ -33,238 +33,398 @@
...
+            //   r in [i, n-1]: read A[r, i]  (column i of stored Lower, rows i..n-1)
+            if (col_major) {
+                blas::axpy(i, av, &A[i], lda, &B[j * ldb], 1);
+                blas::axpy(n - i, av, &A[i + i * lda], 1, &B[i + j * ldb], 1);
+            } else {
+                blas::axpy(i, av, &A[i * lda], 1, &B[j], ldb);
+                blas::axpy(n - i, av, &A[i * lda + i], lda, &B[i * ldb + j], ldb);
+            }
+        }
+    }
```
rileyjmurray: This is a COOMatrix kernel. It should live in RandBLAS/sparse_data.

RandBLAS/sksy.hh
```diff
@@ -33,238 +33,398 @@
...
 ///       * If zero, then :math:`A` is not accessed.
 ///
-///      S - [in]  
-///       * A DenseSkOp or SparseSkOp object.
-///       * Defines :math:`\submat(\mtxS).`
+///      S - [in]
+///       * A SketchingOperator object (DenseSkOp or SparseSkOp).
+///       * Defines :math:`\submat(\mtxS).` SparseSkOp is currently not supported and
+///         calls with a SparseSkOp will throw a :math:`\ttt{RandBLAS::Error}` ---
+///         this is the Case-B kernel from `project-plans/randblas-symm-plan.md`.
```
rileyjmurray: So this actually reduces RandBLAS' functionality. Before this PR we'd be able to sketch a dense structurally-symmetric matrix with a SparseSkOp. After this PR we'd throw an error. That's a "move to RandBLAS 2.0" kind of change. I don't want to make that move in this PR. At the very least this PR has a bunch of stuff that's perfectly useful for RandBLAS 1.x.

RandBLAS/sksy.hh
```diff
@@ -278,255 +438,227 @@ inline void sketch_symmetric(
...
+    blas::Layout layout, blas::Uplo uplo,
+    int64_t d, int64_t n,
+    T alpha,
+    const SparseSkOp<T, RNG, sint_t> &S, int64_t ro_s, int64_t co_s,
+    const T* A, int64_t lda,
+    T beta,
+    T* B, int64_t ldb
+) {
+    RandBLAS::sparse::lsksys(layout, uplo, d, n, alpha, S, ro_s, co_s, A, lda, beta, B, ldb);
 }
```
rileyjmurray: If this PR changes things so that LSKSYS errors on a SparseSkOp S, then we should just have a static_assert(false) here.

rtd/source/FAQ.rst
```diff
@@ -110,10 +110,16 @@ No support for negative values of "incx" or "incy" in sketch_vector.
...
+sketch_symmetric supports both DenseSkOp and SparseSkOp.
+  ``sketch_symmetric`` now takes a ``blas::Uplo`` and exploits symmetry of :math:`A`:
+  the ``DenseSkOp`` branch dispatches to ``blas::symm`` (via ``lsksy3`` / ``rsksy3``)
+  and the ``SparseSkOp`` branch dispatches to a hand-rolled kernel (``lsksys`` /
+  ``rsksys``) that emits two ``blas::axpy`` calls per stored nonzero of the SkOp,
+  reading only the triangle of :math:`A` named by ``uplo``. See
+  ``RandBLAS/sparse_data/DevNotes.md`` for the access-pattern detail.
+
+Layout-mismatched ``DenseSkOp`` in ``sketch_symmetric`` falls back to GEMM.
+  When the ``DenseSkOp``'s storage layout differs from the caller's ``layout`` parameter, ``sketch_symmetric`` falls back to ``blas::gemm`` with the transpose flag --- ``blas::symm`` has no on-the-fly transpose flag for the dense operand. The layout-matched case still gets the SYMM speedup.
```
rileyjmurray: I don't think this is true.

rileyjmurray (changes requested):

rtd/source/api_reference/sketch_sparse.rst
mmelnich: Yes, correct. That would be the only dangerous case for spmm/spgemm application path.

RandBLAS/sksy.hh
mmelnich: Small local tests, but we should not mention it.
Removed

RandBLAS/sksy.hh
mmelnich: Done.

RandBLAS/sksy.hh
mmelnich: Done.

rtd/source/FAQ.rst
mmelnich: Yea, that was stale. Fixed.

RandBLAS/sksy.hh
mmelnich: This isn't a stub. Case B is implemented in this PR (commit 77a3b45). lsksys is a real kernel doing the per-stored-nonzero two-AXPY scatter against dense symm A; the wrapper just forwards into it. After the refactor on this branch it'll call into RandBLAS::sparse_data::coo_lsksys instead, but it's still a real kernel with no error path.

RandBLAS/sksy.hh
mmelnich: Stale docstring, sorry. Case B is actually implemented in this PR (commit 77a3b45): SparseSkOp dispatches to a hand-rolled lsksys / rsksys two-AXPY scatter. No throw, no functionality removed. Fixing the docstring in the next commit.

RandBLAS/sparse_data/DevNotes.md
mmelnich: But we already have sparse * sparse through MKL.

mmelnich: Per the update from just now, msg from Robot friend:
Case D no longer carries a sparse-times-sparse implementation of its own. RandBLAS already exposes sparse-times-sparse as spgemm (MKL-backed), and the Case D overload now routes through exactly that path: it expands A's stored triangle into a general sparse matrix in O(nnz) memory (expand_symmetric_to_general, needed because MKL rejects SYMMETRIC descriptors in sp2m/spmmd) and calls the existing spgemm-to-dense routine with a GENERAL descriptor. B stays sparse. The only Case-D-specific code left is that expansion helper plus dispatch; the old densify-B composition survives only as the fallback for non-MKL builds, where spgemm does not exist.

mmelnich:
@rileyjmurray I kind of forgot about this PR. Could you take a look again?

RandBLAS/sparse_data/coo_sksys_impl.hh
```diff
@@ -0,0 +1,150 @@
...
+            int64_t r = col_S - co_s;  // row of sym(A) contributing to column c
+            // sym(A, uplo)(r, c): read (r, c) if it lies in the stored
+            // triangle, else the mirrored (c, r).
+            bool mirrored = upper ? (r > c) : (r < c);
+            T a_rc = mirrored ? A[c * irs_a + r * ics_a]
+                              : A[r * irs_a + c * ics_a];
+            B_c[i * irs_b] += alpha * Scoo.vals[p] * a_rc;
+        }
+    }
+}
```
rileyjmurray, line 105: Why does this have its own implementation? Seems it could be handled by packing the sketching operator into a COOMatrix and then calling coo_spsymm.

RandBLAS/sparse_data/mkl_spmm_impl.hh
```diff
@@ -256,6 +257,38 @@ MKLSparseHandle make_mkl_handle(const SpMat& A) {
...
+            op, alpha, A_handle, descr, mkl_layout,
+            B, (MKL_INT)n_rhs, (MKL_INT)ldb,
+            beta, C, (MKL_INT)ldc
+        );
+    } else {
+        static_assert(sizeof(T) == 0, "MKL sparse BLAS only supports float and double.");
+        // see GitHub PR #155 for why we don't use static_assert(false, ...).
+    }
+}
+
```
rileyjmurray, line 292: Good stuff!

RandBLAS/sparse_data/mkl_spmm_impl.hh
```diff
@@ -383,20 +405,8 @@ void mkl_spgemm_to_dense(
...
-                for (int64_t j = 0; j < n; ++j)
-                    blas::scal(m, beta, &C[j * ldc], 1);
-            } else {
-                for (int64_t i = 0; i < m; ++i)
-                    blas::scal(n, beta, &C[i * ldc], 1);
-            }
-        }
+        // Just scale C by beta.
+        RandBLAS::util::lascl(layout, m, n, beta, C, ldc);
         return;
```
rileyjmurray, line 418: <3

rtd/source/api_reference/sketch_dense.rst
```diff
@@ -67,29 +67,37 @@ Analogs to GEMM
...
 .. dropdown:: Variants using  :math:`\submat(\mtxS)`
     :animate: fade-in-slide-down
     :color: light
 
-    .. doxygenfunction:: RandBLAS::sketch_symmetric(blas::Layout layout, int64_t d, int64_t n, T alpha, const SKOP &S, int64_t ro_s, int64_t co_s, const T *A, int64_t lda, T beta, T *B, int64_t ldb, T sym_check_tol = 0)
+    .. doxygenfunction:: RandBLAS::sketch_symmetric(blas::Layout layout, blas::Uplo uplo, int64_t d, int64_t n, T alpha, const SKOP &S, int64_t ro_s, int64_t co_s, const T *A, int64_t lda, T beta, T *B, int64_t ldb)
       :project: RandBLAS
 
-    .. doxygenfunction:: RandBLAS::sketch_symmetric(blas::Layout layout, int64_t n, int64_t d, T alpha, const T *A, int64_t lda, const SKOP &S, int64_t ro_s, int64_t co_s, T beta, T *B, int64_t ldb, T sym_check_tol = 0)
+    .. doxygenfunction:: RandBLAS::sketch_symmetric(blas::Layout layout, blas::Uplo uplo, int64_t n, int64_t d, T alpha, const T *A, int64_t lda, const SKOP &S, int64_t ro_s, int64_t co_s, T beta, T *B, int64_t ldb)
```
rileyjmurray, line 96: Clarify. Is it the case that this PR removes support for the old function calls (no `uplo`, explicit `sym_check_tol`), or are we just no longer advertising them on the web docs?

RandBLAS/sparse_data/mkl_spmm_impl.hh
mmelnich: YAY

rtd/source/api_reference/sketch_dense.rst
mmelnich: Removes outright.

RandBLAS/sparse_data/coo_sksys_impl.hh
mmelnich: Looks like that is not the right logic.

Answer per the robot:
 In coo_spsymm the sparse matrix is the symmetric one and the dense matrix is general. Here it is the reverse: the sparse matrix is the sketching operator (rectangular, not symmetric) and the symmetric matrix is the dense one. Handing S to coo_spsymm would make it treat S as the symmetric matrix, which it is not (and coo_spsymm requires its sparse input to be square). The two kernels do share the same loop design; they just read the stored triangle from different storage (sparse scan vs dense indexing).

mmelnich:
@rileyjmurray Anything else you'd like me to do here?

rileyjmurray:
Please resolve merge conflicts. I stole the `lascl` utility in a PR I just merged. After that, please revert any API-breaking changes. You can extend the API with function overloading as needed.

Please have an agent do a comprehensive review of the PR with reference to `CONTRIBUTORS.md`, `AGENTS.md`, and `STYLE_GUIDE.md`. On the topic of style, have it make inferences or draw comparisons to the PR I just merged: https://github.com/BallisticLA/RandBLAS/commit/49099b84f8818dfee4893f44eb41e043ba192314.

mmelnich:
@rileyjmurray how's this looking?

rtd/source/FAQ.rst
```diff
@@ -110,10 +110,15 @@ No support for negative values of "incx" or "incy" in sketch_vector.
...
-  in sketch_general to resolve layout discrepancies between the various arguments.
+sketch_symmetric supports both DenseSkOp and SparseSkOp.
+  ``sketch_symmetric`` now takes a ``blas::Uplo`` and exploits symmetry of :math:`A`:
+  the ``DenseSkOp`` branch dispatches to ``blas::symm`` (via ``lsksy3`` / ``rsksy3``)
+  and the ``SparseSkOp`` branch dispatches to a column-driven accumulation kernel
+  (``lsksys`` / ``rsksys``) that reads only the triangle of :math:`A` named by
+  ``uplo``. See ``RandBLAS/sparse_data/DevNotes.md`` for the access-pattern detail.
+
+Layout-mismatched ``DenseSkOp`` in ``sketch_symmetric`` transpose-copies the operand.
+  When the ``DenseSkOp``'s storage layout differs from the caller's ``layout`` parameter, ``sketch_symmetric`` transpose-copies the operand into a tight buffer in the caller's layout and then calls ``blas::symm``. The copy is ``O(d * n)``; ``blas::symm`` has no on-the-fly transpose flag for the dense operand, so this is what it costs to keep the SYMM speedup over a ``blas::gemm`` fallback. The layout-matched case skips the copy.
```
rileyjmurray: This material is under a **Limitations** heading, but not written in a way that describes limitations. It's just a very detailed behavior explanation of a certain codepath.Please resolve. Have the robot read this file in its entirety so it keeps style consistent.

Note: sketching operator layout-mismatches are already resolved by copy-transpose, and we don't mention that on this page; see fill_dense_unpacked.

rtd/source/api_reference/sketch_dense.rst
rileyjmurray: > Removes outright.

No longer true. Marking as resolved.

rtd/source/api_reference/sketch_dense.rst
```diff
@@ -67,25 +67,55 @@ Analogs to GEMM
 Analogs to SYMM
 ---------------
 
+These overloads accept a ``blas::Uplo`` parameter naming the triangle of
+:math:`\mtxA` that is structurally stored; the opposite triangle is implied
+by symmetry and is **not** read. Both ``DenseSkOp`` and ``SparseSkOp`` are
+supported: DenseSkOp dispatches to ``blas::symm`` via ``lsksy3`` / ``rsksy3``,
+SparseSkOp dispatches to a column-driven accumulation kernel via
+``lsksys`` / ``rsksys``. See ``RandBLAS/sparse_data/DevNotes.md`` for the
+SparseSkOp access-pattern detail.
```
rileyjmurray: This level of detail isn't present in similar pages.

rtd/source/api_reference/sketch_sparse.rst
```diff
@@ -114,6 +114,55 @@ Deterministic operations
...
+       covering all 3 x 3 = 9 sparse-format pairings for ``(A, B)``.
+       ``mkl_sparse_sp2m`` rejects symmetric descriptors and
+       ``mkl_sparse_?_spmmd`` takes no descriptor, so the symmetric
+       expansion happens on the RandBLAS side: on MKL builds the stored
+       triangle of ``A`` is expanded to a general sparse matrix in
+       ``O(nnz)`` memory and the existing sparse-times-sparse routine runs
+       with a GENERAL descriptor, keeping ``B`` sparse. Non-MKL builds
+       (and index widths mismatched with ``MKL_INT``) fall back to
+       densifying ``B`` into a temporary ``m``-by-``n`` buffer and
+       composing through the Case-C dispatcher.
```
rileyjmurray: References to Cases B, C, D aren't appropriate here, since those aren't defined on this page.

Separately --- comments at the kernel-dispatch level don't belong on the website.

rtd/source/api_reference/sketch_sparse.rst
```diff
@@ -114,6 +114,55 @@ Deterministic operations
...
 
+.. dropdown:: :math:`\mtxY = \alpha \cdot \mtxA \cdot \mtxB + \beta \cdot \mtxY,` with sparse symmetric :math:`\mtxA` (spsymm)
+    :animate: fade-in-slide-down
+    :color: light
+
+    .. doxygenfunction:: RandBLAS::spsymm(blas::Layout layout, blas::Uplo uplo, int64_t m, int64_t n, T alpha, const SpMat &A, const T *B, int64_t ldb, T beta, T *Y, int64_t ldy)
+      :project: RandBLAS
+
+    .. doxygenfunction:: RandBLAS::spsymm(blas::Layout layout, int64_t m, int64_t n, T alpha, const Symmetric<SpMat> &A_sym, const T *B, int64_t ldb, T beta, T *Y, int64_t ldy)
+      :project: RandBLAS
```
rileyjmurray: Variable name comment: name the matrices (A, B, C) or (A, X, Y). The current approach (A, B, Y) is no good. Apply appropriate changes everywhere that makes sense -- source and docs.

rtd/source/api_reference/sketch_sparse.rst
rileyjmurray: Don't restrict template to `DenseSkOp`.

rtd/source/api_reference/sketch_sparse.rst
```diff
@@ -114,6 +114,55 @@ Deterministic operations
        This function requires Intel MKL and only supports single and double precision (``float`` and ``double``),
        in contrast to other RandBLAS kernels that work with any scalar type.
 
+.. dropdown:: :math:`\mtxY = \alpha \cdot \mtxA \cdot \mtxB + \beta \cdot \mtxY,` with sparse symmetric :math:`\mtxA` (spsymm)
+    :animate: fade-in-slide-down
+    :color: light
+
+    .. doxygenfunction:: RandBLAS::spsymm(blas::Layout layout, blas::Uplo uplo, int64_t m, int64_t n, T alpha, const SpMat &A, const T *B, int64_t ldb, T beta, T *Y, int64_t ldy)
+      :project: RandBLAS
```
rileyjmurray: Does MKL have two dimensional parameters (`m` and `n`) in its SPMM function? I think it'd only have one.

rtd/source/api_reference/sketch_sparse.rst
rileyjmurray: If we're going use function overloading here then the typed `Symmetric<SpMat>` argument should probably be in an overload of `spmm`, not `spsymm`.

rileyjmurray (changes requested):
There  are lots of design questions here. These are just some initial thoughts.

rtd/source/api_reference/sketch_dense.rst
mmelnich: Trimmed

rtd/source/api_reference/sketch_sparse.rst
mmelnich: Removed

rtd/source/api_reference/sketch_sparse.rst
mmelnich: Renamed to (A, B, C) everywhere

rtd/source/api_reference/sketch_sparse.rst
mmelnich: You're right. Fixed.

rtd/source/api_reference/sketch_sparse.rst
mmelnich: Moved, now spmm(n, alpha, Symmetric<SpMat>, ...), same overload set as general spmm.

rtd/source/api_reference/sketch_sparse.rst
mmelnich: Fixed.

mmelnich:
@rileyjmurray addressed ur comments

mmelnich:
@rileyjmurray any progress on this?

rileyjmurray:
No immediate progress. Is this blocking a paper?

mmelnich:
> No immediate progress. Is this blocking a paper?

It will be in a bit over 2 weeks

rileyjmurray:
Understood

PR #176, opened 2026-06-19
Sparse sketching speedups

rileyjmurray:
This branch reworks how COO-represented sparse sketching operators are applied to
dense matrices in `lskges`/`rskges`. Instead of always converting to CSC and running a
single kernel, we now choose a compressed format (CSR or CSC) per call and hand the
operator to the same per-RHS-column kernels that `left_spmm`/`right_spmm` already use
for native CSR/CSC operands. The sampling path was also adjusted so that operators
arrive at the apply step already sorted, which lets the conversion be a zero-copy view
in the common case.

## New kernels and dispatching logic

- **Format-selection heuristic** (`coo_spmm_target_format` in `coo_spmm_impl.hh`).
  The per-column kernels loop the operator's outer dimension once per output column, so
  a shorter outer loop is cheaper. For the ColMajor/ColMajor path with a *wide*
  operator (`d < m`), we now route through the CSR `jik` kernel; every other case stays
  on the CSC kernels. The RowMajor path keeps the bandwidth-saturating CSC `kib` axpy
  kernel.
- **COO fallback rewritten** (`apply_coo_via_csc` → `apply_coo_via_csx`). The COO
  branch of `left_spmm` now materializes the operator in the heuristically-chosen format
  and dispatches to the matching CSR/CSC kernel, rather than hard-coding CSC.
- **`lskges`/`rskges` custom dispatch.** Full operators (sampled or not) go through new
  `_lskges_compress_and_apply_coo` / `_rskges_compress_and_apply_coo` helpers that pick
  the compressed format and apply it. The right-apply helper accounts for `right_spmm`
  reducing to a transposed `left_spmm` (which flips CSR<->CSC) by requesting the opposite
  format. Proper submatrices of an already-sampled operator still fall back to the generic
  `left_spmm`/`right_spmm` offset handling.

## Supporting infrastructure

- **Zero-copy compressed views** (`coo_to_csr_view_or_copy` /
  `coo_to_csc_view_or_copy`). When a COO is already sorted in the target order, we build
  a CSR/CSC that shares the COO's value and index arrays and only allocates the pointer
  array — avoiding a deep copy on the hot apply path. Otherwise they fall back to a full
  conversion.
- **O(nnz) counting-sort transpose** (`counting_sort_transpose`). Converting between the
  two opposing compressed orders (e.g. a CSC-sorted COO into CSR) now uses a counting-sort
  transpose instead of a comparison re-sort. Wired into `coo_to_csr` / `coo_to_csc`.
- **Sampling emits sorted nonzeros.** `fill_sparse_unpacked` now sorts each major-axis
  vector as it is sampled (small no-alloc insertion sort for SASO; post-merge sort for
  LASO). `submatrix_as_coo` labels the result as CSR- or CSC-sorted instead of `None`, so
  the apply path can take the zero-copy view.
- **Small shared helpers**: `flipped_layout` (base.hh), `flipped_nonzerosort`
  (sparse_data/base.hh), and a `restrict_to_window` helper for windowed COO operators.

## Other material changes

- **Removed the "regular CSC" fixed-nnz-per-column kernel** and its runtime probe. It was
  previously found to be no faster (and sometimes slower) than the general kernel, and the
  probe added a scan over the data on every call.
- **New benchmark**: `examples/.../sketch_general_performance.cc` with work-normalized
  metrics and a `--csr-probe` mode used to calibrate the dispatch heuristic.
- **Tests**: new CSR SpMM coverage (`test_spmm_csr.cc`), COO-matrix tests for the
  accumulate semantics, and updated sparse-skop tests for the new sort labeling.
- Renamed the workspace `CLAUDE.md` to `AGENTS.md` and refreshed dev notes.


RandBLAS/sparse_data/coo_spmm_impl.hh
rileyjmurray, line 1: TODOs
  * move counting_sort_transpose to sparse_data/base.hh
  * have LSKGES explicitly instantiate CSR (if column-major) or CSC (if row-major), assuming opS=NoTrans.
  * move the sketching-specific logic out of apply_coo_via_csc.
  * Rename apply_coo_via_csc to reflect that it's just a dispatcher. (Unclear how to frame it.)

RandBLAS/sparse_data/csr_spmm_impl.hh
rileyjmurray, line 1: TODO: remove `regular` kernels. There's hardly any memory traffic savings. The most it could help with is signaling if it's safe to use `static` or `dynamic` openmp threading. The only straightforward way we have of mapping to the regular path with LASOs involves coefficient splitting that increases memory traffic AND floating point cost.

TODO: consider removing `regular` CSC kernels, which we use for SASOs. They don't have a cost like we face with LASOs, but they're likely offering no performance improvement at the cost of increased maintenance.

RandBLAS/sparse_data/DevNotes.md
```diff
@@ -73,3 +73,28 @@ From there, we'll do the following.
...
+ * **SpMM and densify accumulate duplicates.** The CSR/CSC kernels sum ``vals[ell]*v[...]`` over a
+   row/column, so duplicate column/row indices add correctly. ``coo_to_dense`` was made to use
+   ``+=`` (not ``=``) for the same reason. Any new densification path must accumulate, not overwrite.
+ * **Conversions preserve duplicates.** ``coo_to_csr`` / ``coo_to_csc`` and the counting-sort
+   transpose in ``coo_spmm_impl.hh`` keep all ``nnz`` records (no de-duplication), which is what the
+   kernels expect.
+ * **Do not route a regular operator through ``trsm``.** ``trsm_dispatch.hh`` validates structure
+   with ``compressed_indices_are_increasing``, which rejects duplicate indices within a major-axis
+   vector. Triangular solve is unrelated to sketching, so this is not a real constraint in practice,
+   but it is a trap worth noting.
```
rileyjmurray: Remove.

RandBLAS/sparse_skops.hh
```diff
@@ -194,10 +194,18 @@ struct SparseDist {
...
     ///  \math{\mtxx_j} will equal \math{\sqrt{\ell}} with probability 1/2 and
     ///  \math{-\sqrt{\ell}} with probability 1/2.
     ///
+    ///  Note: while \math{\mtxx} has at most \math{\vecnnz} *distinct* nonzero coordinates, it is
+    ///  *stored* with exactly \math{\vecnnz} structural entries per major-axis vector. A coordinate
+    ///  \math{j} sampled \math{\ell} times is held as \math{\ell} duplicate records each equal to
+    ///  \math{\mtxx_j/\ell = \pm 1/\sqrt{\ell}}, which sum to \math{\mtxx_j.} This "regular" layout
+    ///  (a fixed number of records per major-axis vector) lets long-axis-major operators reach the
+    ///  same fixed-nnz fast-path kernels as short-axis-major operators, without changing the operator
+    ///  that \math{\mtxx} represents.
```
rileyjmurray: Remove.

RandBLAS/sparse_skops.hh
```diff
@@ -483,7 +491,8 @@ template <typename T, SignedInteger sint_t>
 void laso_merge_long_axis_vector_coo_data(
     int64_t vec_nnz, T* vals, sint_t* idxs_lax, sint_t *idxs_sax, int64_t i,
     std::unordered_map<sint_t, T> &loc2count,
-    std::unordered_map<sint_t, T> &loc2scale
+    std::unordered_map<sint_t, T> &loc2scale,
+    bool regular
```
rileyjmurray: remove

RandBLAS/sparse_skops.hh
```diff
@@ -505,7 +514,19 @@ void laso_merge_long_axis_vector_coo_data(
...
+        // s/sqrt(c) (s = loc2scale[ell], the canonical sign). The c copies then sum to
+        // c * (s/sqrt(c)) = sqrt(c)*s -- the exact value the merge branch below would
+        // store as a single entry. So the densified operator is identical; we simply
+        // leave duplicate (idxs_lax, idxs_sax) records in place to give a fixed-vec_nnz
+        // "regular" layout (every major-axis vector has exactly vec_nnz records).
+        for (int64_t j = 0; j < vec_nnz; ++j) {
+            sint_t ell = idxs_lax[j];
+            vals[j] = loc2scale[ell] / std::sqrt(loc2count[ell]);
+        }
+    } else if ((int64_t) loc2scale.size() < vec_nnz) {
```
rileyjmurray: remove

test/datastructures/test_coo_matrix.cc
```diff
@@ -64,7 +64,10 @@ void sparseskop_to_dense(
         sint_t row = S0.rows[i];
         sint_t col = S0.cols[i];
         T val = S0.vals[i];
-        mat[idx(row, col)] = val;
+        // Accumulate: a regular long-axis-major operator stores collided coordinates as
+        // duplicate records that must sum (matches coo_to_dense). The buffer is zeroed
+        // above, so += equals = for a deduplicated operator.
```
rileyjmurray: Remove comment. Keep change.

RandBLAS/sparse_data/base.hh
```diff
@@ -300,6 +311,33 @@ void compressed_ptr_to_sorted_idxs(int64_t num_comp, sint_t1* ptr, int64_t len_i
...
+    for (int64_t g = 0; g < dim; ++g)  ptr[g + 1] += ptr[g];
+    std::vector<sint_t> cursor(ptr, ptr + dim);   // running write position per group
+    for (int64_t e = 0; e < nnz; ++e) {
+        sint_t g = group_idx[e];
+        sint_t pos = cursor[g]++;
+        out_other[pos] = other_idx[e];
+        out_vals[pos]  = vals[e];
+    }
+}
+
```
rileyjmurray, line 338: TODO: Rename to `transpose_compressed`. Have docstring note sort-preservation properties. Example properties: CSC -> CSR, CSR -> CSC, and applying twice to degenerate CSR/CSC is equivalent to applying a per-compressed-vector sort (making it proper CSR/CSC).

RandBLAS/sparse_data/coo_spmm_impl.hh
```diff
@@ -49,8 +50,51 @@ using RandBLAS::SignedInteger;
...
+// both SASO and LASO, with no regression -- so we prefer CSR there. The RowMajor path keeps
+// the bandwidth-saturating CSC kib axpy kernel. See the --csr-probe mode of
+// examples/.../sketch_general_performance.cc for measurements.
+inline NonzeroSort coo_spmm_target_format(
+    blas::Layout layout_opB, blas::Layout layout_C, int64_t d, int64_t m
+) {
+    bool col_major  = (layout_opB == blas::Layout::ColMajor && layout_C == blas::Layout::ColMajor);
+    bool prefer_csr = col_major && (d < m);
+    return prefer_csr ? NonzeroSort::CSR : NonzeroSort::CSC;
+}
```
rileyjmurray, line 66: Reinvestigate this; probably define a separate version when sketching.

RandBLAS/sparse_skops.hh
```diff
@@ -658,7 +679,21 @@ state_t fill_sparse_unpacked(
...
+                sint_t key = im[a];
+                T      vv  = v[a];
+                int64_t c = a - 1;
+                for (; c >= 0 && im[c] > key; --c) {
+                    im[c+1] = im[c];
+                    v[c+1]  = v[c];
+                }
+                im[c+1] = key;
+                v[c+1]  = vv;
+            }
```
rileyjmurray: Probably good to consolidate insertion-sorts into a helper function.

RandBLAS/sparse_skops.hh
```diff
@@ -849,7 +884,10 @@ COOMatrix<T, sint_t> submatrix_as_coo(
     A.rows = rows;
     A.cols = cols;
     A.nnz  = nnz;
-    A.sort = RandBLAS::sparse_data::NonzeroSort::None;
+    // fill_sparse_unpacked emits each major-axis vector in sorted order, so the sampled
+    // submatrix is CSR- or CSC-sorted; label it as such (rather than None) so downstream
+    // conversions can take an O(nnz) fast path.
+    A.sort = RandBLAS::sparse_data::coo_arrays_determine_sort(A.nnz, A.rows, A.cols);
```
rileyjmurray: I'm pretty sure this is an unnecessary scan of the matrix.

AGENTS.md
rileyjmurray, line 1: Replace "codex" with "coding agent(s)"

RandBLAS/sparse_data/base.hh
rileyjmurray: Will handle that TODO in another PR. For now I've just slimmed down the comments.

RandBLAS/sparse_skops.hh
rileyjmurray: Done

PR #179, opened 2026-07-21
Windows portability

RaphaelArkadyMeyerNYU:
This pull should provide basic windows (MSVC) compatibility.

I got this running on a Surface Book 3 running Windows x64, MSVC 19.51, C++20.

The exact install information how to run this on Windows is shown in the install.md file.

Several changes had to be made to make this work. An LLM-Generated summary of the work is attached below.

---

# Native Windows portability notes

This document records the issues encountered while configuring RandBLAS natively on Windows with MSVC, CMake, NMake, standalone Intel oneMKL, BLAS++, and Random123.

Last fully audited against the working native-MSVC build and Windows CI:
2026-07-24.

### 1. Fixed: installed RandBLAS config embedded Windows paths incorrectly

`CMake/RandBLASConfig.cmake.in` writes dependency paths directly into generated CMake code:

```cmake
set(blaspp_DIR @blaspp_DIR@)
set(Random123_DIR @Random123_DIR@)
```

When configured with ordinary Windows backslashes, a generated value such as
`C:\Users\...` contains CMake escape sequences (`\U`) and causes downstream
`find_package(RandBLAS)` to fail with `Invalid character escape '\U'`.

The current branch normalizes both dependency directories with
`file(TO_CMAKE_PATH ...)` before substitution and quotes the generated values.
The Windows downstream-consumer job deliberately configures RandBLAS with
native-backslash dependency paths and then loads the installed package without
re-supplying them, serving as a regression test for this fix.

The fix has been verified locally end to end with MSVC: RandBLAS was configured
using native-backslash dependency paths, installed into a temporary prefix,
and consumed using only that prefix. The downstream smoke target configured,
compiled, linked, and ran successfully. The generated config contained
normalized `C:/...` values. The subsequent GitHub-hosted downstream run also
passed.

### 2. Fixed: compiler flags use toolchain guards

The original `RandBLAS/CMakeLists.txt` added `-Wall`, `-Wextra`, and
`-Wno-comment` through the header-only interface target without a compiler
check. MSVC does not use these options. The current branch keeps that warning
set (including `-Wno-comment`) for GNU/Clang and uses `/W4` for MSVC.

The `SANITIZE_ADDRESS` path is now compiler-aware as well. GNU and Clang receive
`-fsanitize=address` at compilation and linking. MSVC receives
`/fsanitize=address` and `/Zi` as compile options, plus `/DEBUG` while linking,
after CMake verifies that the complete compiler/runtime setup supports it.
Unsupported compilers, or an MSVC
installation without the optional AddressSanitizer runtime libraries, fail
during configuration with a targeted explanation.

Raphael's current MSVC installation recognizes the switch but lacks
`clang_rt.asan_dynamic_runtime_thunk-x86_64.lib`, so the capability check
correctly rejects `-DSANITIZE_ADDRESS=ON`. A normal non-sanitized build remains
unaffected.

### 3. Fixed: examples included the POSIX-only `<unistd.h>` header

The following examples included `<unistd.h>` even though the header was not
used:

* `examples/total-least-squares/tls_dense_skop.cc`
* `examples/total-least-squares/tls_sparse_skop.cc`
* `examples/sparse-low-rank-approx/qrcp_matrixmarket.cc`
* `examples/sparse-low-rank-approx/svd_rank1_plus_noise.cc`
* `examples/sparse-low-rank-approx/svd_matrixmarket.cc`

The current branch removes all five unused includes. If a POSIX API becomes
necessary later, its include should instead be guarded by platform detection.

### 4. Open design question: example CMake requires OpenMP unconditionally

`examples/CMakeLists.txt` uses `find_package(OpenMP REQUIRED)`, whereas the core
RandBLAS build supports serial operation. Native MSVC OpenMP is now validated:
the installed RandBLAS package changes the imported OpenMP target to
`/openmp:experimental`, and the examples compile and run. The remaining design
question is whether examples should share the core optional-OpenMP behavior or
explicitly document OpenMP as a requirement.

### 5. Fixed: total-least-squares examples used variable-length arrays

The TLS examples declared stack arrays whose sizes depend on runtime values:

```cpp
double target_x[n*d];
double eps[m*d];
```

Variable-length arrays are not part of standard C++ and MSVC rejects them with
error C2131. The current branch replaces these arrays with
`std::vector<double>` and passes their storage through `.data()` to RandBLAS and
BLAS++.

### 6. Partially fixed: documentation and CI were Unix-centric

`INSTALL.md` and the examples documentation use `make`, shell backticks, and
Unix path conventions. `RAPHAEL_INSTRUCTION.md` now records the validated local
NMake recipe, but the project still has no general Windows installation guide.

The current branch adds native-MSVC jobs to the core, downstream-consumer, and
examples workflows. A Windows composite action provisions oneMKL through
vcpkg, then builds the remaining dependencies. A shared PowerShell driver makes
the substantive configure, build, test, and example commands runnable both in
GitHub Actions and on a local x64 Native Tools installation. Current results
are summarized under **Validation and CI status**.

The documentation and thread-sanitizer workflows remain Unix-only
intentionally. Documentation generation is platform-independent and receives
little additional signal from duplicating it on Windows. The available
Clang/GCC thread-sanitizer setup is not a native-MSVC facility. The CMake
AddressSanitizer flags are now compiler-aware; adding a Windows
AddressSanitizer job remains optional and would require a runner/toolchain with
the MSVC AddressSanitizer runtime installed.

### 7. Fixed for RandBLAS: upstream dependency portability remains

The downstream smoke test exposed additional portability issues:

* The current Random123 `boxmuller.hpp` fallback does not enable
  `R123_NO_SINCOS` for MSVC, so `sincosf` and `sincos` are undeclared. RandBLAS
  already supplies host `sincospif`/`sincospi` shims, but the Random123 fallback
  also needs to be enabled (or equivalent MSVC-safe wrappers need to be supplied).
* Upstream BLAS++ `include/blas/symv.hh` contains an unconditional diagnostic
  using `__PRETTY_FUNCTION__`, which MSVC does not define. The selected
  `windows-portability` branch removes that diagnostic.

The same build reported parser errors in the RandBLAS `uneg11` wrapper because
MSVC's legacy `__cplusplus` value hid Random123's C++11 templates. Supplying
`/Zc:__cplusplus` through the RandBLAS target resolved those errors; they were
not an independent RandBLAS parser defect.

The current branch exports `R123_NO_SINCOS=1` as an MSVC-only `INTERFACE`
definition on the `RandBLAS` target. This provides a scoped downstream fix
without modifying the Random123 checkout. Random123 should still detect MSVC
itself upstream so direct Random123 consumers receive the same fallback.
RandBLAS's test diagnostics now use the descriptive
`__RANDBLAS_PRETTY_FUNCTION__` macro from
`RandBLAS/testing/comparison.hh`. It expands to `__FUNCSIG__` under MSVC,
`__PRETTY_FUNCTION__` under GNU/Clang, and the standard `__func__` fallback
elsewhere. This keeps the compiler extension inside the test helper and removes
the former `__PRETTY_FUNCTION__=__FUNCSIG__` definition from the exported
RandBLAS target.

### 8. Fixed: MSVC must report the selected standard in `__cplusplus`

MSVC historically reports `__cplusplus` as `199711L` even when compiling with
`/std:c++20`. Random123 uses `__cplusplus >= 201103L` to expose C++11 facilities
such as `uneg11all`. Without `/Zc:__cplusplus`, the template is hidden and
RandBLAS later fails while parsing `r123::uneg11all<T>`.

The current branch adds `/Zc:__cplusplus` and `/EHsc` as compiler-conditioned
`INTERFACE` options on the header-only `RandBLAS` CMake target. They therefore
apply both to the source-tree tests and to consumers of the installed package,
without affecting non-MSVC compilers. Together with the MSVC-only compile
definition for Random123, RandBLAS-related configure commands no
longer require custom `CMAKE_CXX_FLAGS`.

Exporting `/EHsc` from the target also preserves normal C++ exception-unwind
semantics if a caller replaces CMake's default `CMAKE_CXX_FLAGS` for unrelated
reasons.

### 9. Fixed for MSVC: `omp simd` requires experimental OpenMP mode

The full test build reaches `RandBLAS/sparse_data/csr_spmm_impl.hh`, which uses
`#pragma omp simd`. MSVC reports:

```text
error C7660: 'simd': requires '-openmp:experimental' command line option(s)
```

The current branch applies `/openmp:experimental` automatically when MSVC is
detected, both while building RandBLAS from source (`CMake/OpenMP.cmake`) and
when an installed package is loaded (`CMake/RandBLASConfig.cmake.in`). No
`OpenMP_CXX_FLAGS` command-line setting is required. Other compiler families
retain CMake's normal OpenMP discovery.

This is compiler-specific rather than a universal Windows requirement.
GCC/MinGW, Clang, and Intel compilers use different OpenMP options and runtimes.
A more general future implementation could probe support for `omp simd` and
fall back to an ordinary correctness-preserving loop when SIMD OpenMP is
unavailable, rather than selecting solely by the `MSVC` compiler ID.

### 10. Resolved configuration: full tests require ILP64 BLAS++

An LP64 BLAS++ build reaches the sparse GEMM tests and then fails when they
instantiate 64-bit (`int64_t`) sparse indices:

```text
static assertion failed: 'RandBLAS::spgemm: first sparse matrix index type must match MKL_INT size.'
```

With oneMKL LP64 (`mkl_intel_lp64_dll.lib`),
`MKL_INT` is 32-bit. RandBLAS's MKL sparse backend requires sparse index types
to have the same size as `MKL_INT`; the tests use 64-bit indices. This is a
configuration mismatch, not an MSVC parser failure.

The canonical Windows recipe therefore uses one ILP64 BLAS++ installation
(`mkl_intel_ilp64_dll.lib`, `blas_int=ilp64`) for tests, installation, smoke
validation, and examples. A separate LP64 installation is unnecessary for this
workflow, though LP64 remains valid for consumers using 32-bit BLAS indices.

MSVC also warns that the `reduction` clause on the SIMD directive is ignored
and reports that the loop was not vectorized. These are performance warnings;
the test build proceeds, but the MSVC SIMD path is not currently providing the
same optimization as GCC/Clang. Follow-up work should check compiler
vectorization reports (and, if relevant, benchmarks) on the Linux and Apple
builds as well, and ask the maintainers whether the CSR loop is expected to
vectorize there. The MSVC warning alone does not establish that the non-Windows
builds are vectorizing successfully.

### 11. Fixed: imported runtime DLLs were unavailable during test discovery

GoogleTest discovery executes each test binary immediately after linking. If
the Windows loader cannot find a dependency, the build fails with exit code
`0xc0000135` even though compilation and linking succeeded.

The Windows-only `randblas_stage_runtime_dlls` helper now copies
`$<TARGET_RUNTIME_DLLS:target>` beside every repository-owned test,
downstream-smoke, and example executable. Test targets invoke the helper before
`gtest_discover_tests`, so the DLLs are present before post-build discovery.
The transitive CMake target data selects the configured BLAS++ and LAPACK++
DLLs without hardcoded filenames or installation paths.

`TARGET_RUNTIME_DLLS` requires CMake 3.21, so native Windows configurations now
reject older CMake releases with a clear diagnostic. The local recipe uses
`cmake --fresh`, which separately requires CMake 3.24; the validated machine
has CMake 3.29.2. Non-Windows builds retain the project's earlier minimum.

oneMKL remains separate: BLAS++ exports MKL as raw `.lib` paths rather than
imported shared-library targets, so CMake cannot infer the corresponding MKL
DLLs. Continue activating the installer-provided oneMKL environment before
configuring, building, or running. No additional RandBLAS runtime `PATH`
wrapper is required. The helper serves repository-owned build-tree executables;
it does not deploy dependencies for arbitrary installed-package consumers.

## Selected dependency revisions

The Windows jobs use vcpkg's `intel-mkl:x64-windows` port and build shared,
sequential, ILP64 BLAS++ against its raw oneMKL libraries. Random123 remains a
header-only dependency. Until the corresponding changes are upstream, CI uses:

* `RaphaelArkadyMeyerNYU/blaspp`, branch `windows-portability`;
* `RaphaelArkadyMeyerNYU/lapackpp`, branch `msvc-compatibility`.

These branches contain the BLAS++ diagnostic removal and LAPACK++ direct-include
fixes. The Actions setup performs ordinary clone, configure, build, and install
steps; it does not rewrite dependency sources or inject compatibility macros.

## Validation and CI status

The Windows workflow coverage is:

* `core.yml`: MSVC/oneMKL ILP64 Release builds with OpenMP enabled and
  disabled, plus a serial AddressSanitizer build. The OpenMP job runs the suite
  with one and four threads; the sanitizer job configures through
  `-DSANITIZE_ADDRESS=ON` and runs the full serial suite under MSVC ASan.
* `downstream-consumer.yml`: installs RandBLAS, then configures, builds, and
  runs the smoke consumer without re-supplying `blaspp_DIR` or
  `Random123_DIR`.
* `examples.yml`: builds LAPACK++, RandBLAS, and all seven examples. Like the
  Linux examples job, it does not execute them.

The previously hosted core, examples, and downstream jobs pass, including clean
vcpkg provisioning. The first Windows AddressSanitizer run reached the link
step and exposed an MSVC STL-annotation mismatch: RandBLAS was instrumented,
but the cached GoogleTest libraries were not. The dependency action now gives
the sanitizer job a separately instrumented and separately cached GoogleTest
installation; that correction is pending its hosted run. The downstream job
permanently exercises native-backslash configure/install/consume behavior and
runs `smoke.exe`.

Additional local validation includes 459/459 OpenMP tests, 455/455 serial tests
(the other four are OpenMP-specific), and manual execution of all seven
examples. The file-based examples used a temporary 80-by-80 Matrix Market
input. The Actions workflows intentionally stop after building the examples.

The composite dependency action is
`.github/actions/setup-randblas-deps-windows/action.yml`; its shared local and
hosted driver is `.github/scripts/windows/run-ci.ps1`. The workflows pass
`actionlint`, YAML parsing, and PowerShell parsing locally.

## Appendix: GitHub Issues

The following drafts capture the remaining upstream work and the RandBLAS
memory-safety fix. They are intentionally self-contained so they can be filed
with minimal editing; historical reproduction flags in an issue body are not
part of the current RandBLAS build recipe.

### A. Random123: MSVC host compilation fails in `boxmuller.hpp`

**Title:** MSVC compilation fails because `boxmuller.hpp` does not enable the `sincos` fallback

**Body:**

When Random123 is included in a native Windows/MSVC C++20 project, compiling
`Random123/boxmuller.hpp` fails with:

```text
error C3861: 'sincosf': identifier not found
error C3861: 'sincos': identifier not found
```

The header provides fallback definitions guarded by `R123_NO_SINCOS`, but the
macro is enabled automatically only for `__APPLE__`. MSVC does not provide the
POSIX `sincos`/`sincosf` functions, so the fallback is also needed there.

Suggested fix: enable `R123_NO_SINCOS` for MSVC (or for host compilers known not
to provide these functions), or use a portable compiler-specific wrapper based
on `std::sin` and `std::cos`. A minimal downstream workaround is to compile
with `R123_NO_SINCOS=1`.

Environment tested: Windows x64, MSVC 19.51, C++20, Random123 headers from the
current repository, standalone Intel oneMKL/BLAS++ consumer.

### B. BLAS++: MSVC compilation fails on `__PRETTY_FUNCTION__`

**Title:** `blas/symv.hh` uses the nonportable `__PRETTY_FUNCTION__` macro on MSVC

**Body:**

Native MSVC compilation of a downstream BLAS++ consumer fails in
`include/blas/symv.hh`:

```text
error C3861: '__PRETTY_FUNCTION__': identifier not found
error C2065: '__PRETTY_FUNCTION__': undeclared identifier
```

The function currently contains:

```cpp
printf("%s: %s\n", __func__, __PRETTY_FUNCTION__);
```

`__PRETTY_FUNCTION__` is a GCC/Clang extension. MSVC provides `__FUNCSIG__` as
its closest equivalent. Since this appears to be an unconditional diagnostic
print in a public header, the preferred fix may simply be to remove the debug
print. Otherwise, BLAS++ should define a portable function-signature macro and
map it to `__FUNCSIG__` on MSVC.

Environment tested: Windows x64, MSVC 19.51, C++20, BLAS++ built with NMake,
standalone Intel oneMKL, downstream RandBLAS smoke test.

### C. LAPACK++: missing direct includes break the native MSVC build

**Title:** MSVC build fails because `config.h` and `lartg.cc` rely on transitive standard-library includes

**Body:**

Building LAPACK++ natively with MSVC fails in the ILP64 configuration with
diagnostics beginning with:

```text
include/lapack/config.h(20): error C4430: missing type specifier
include/lapack/config.h(20): error C2146: missing ';' before identifier 'lapack_int'
include/lapack/fortran.h(35): error C2065: 'LAPACK_S_SELECT2': undeclared identifier
src/lartg.cc(36): error C2039: 'complex': is not a member of 'std'
```

There appear to be two independent missing direct includes:

1. `include/lapack/config.h` uses global `int64_t` when `LAPACK_ILP64` is
   enabled, but does not include `<stdint.h>`. Some Unix standard-library
   headers expose this type transitively; MSVC's `<stdlib.h>` does not.
2. `src/lartg.cc` uses `std::complex` but does not include `<complex>`. Under
   MSVC, the `_MSC_VER` branch in `lapack/config.h` defines C-compatible complex
   structs and does not include the C++ `<complex>` header.

The `LAPACK_S_SELECT2` and related callback errors appear to cascade from the
failed `lapack_int` typedef. The minimal source fixes are:

```cpp
// include/lapack/config.h
#include <stdint.h>
```

```cpp
// src/lartg.cc
#include <complex>
```

These headers should be included directly rather than relying on transitive
implementation details of a particular C or C++ standard library.

Environment tested: Windows x64, MSVC 19.51, C++17, LAPACK++ version
2025.05.28 (commit `8b32767`), NMake generator, BLAS++/oneMKL ILP64 sequential
configuration.

### D. RandBLAS: off-by-`k` workspace offset corrupts the heap in sparse SVD examples

**Title:** Sparse SVD examples write past the `qb_to_svd` workspace allocation

**Body:**

Running the synthetic sparse low-rank approximation example can terminate with
heap corruption during the conversion of a QB factorization to an SVD. A
native Windows/MSVC reproduction is:

```text
slra_svd_synthetic.exe 200 100 4
16
13333
13338
Exit code 0xc0000374
```

The `qb_to_svd` implementations in both
`examples/sparse-low-rank-approx/svd_rank1_plus_noise.cc` and
`examples/sparse-low-rank-approx/svd_matrixmarket.cc` divide a caller-provided
workspace into two regions. The first `k*k` elements hold `U`, and the next
`m*k` elements hold a temporary copy of `Q`. The available second-region size
is correspondingly calculated as:

```cpp
int64_t extra_work_size = lwork - k*k;
```

However, the second region begins at `work + k*(k+1)`. When the caller supplies
the documented/expected `k*k + m*k` elements, only `m*k - k` elements remain
at that address, and `lapack::lacpy` writes `k` elements beyond the allocation.
For the reproduction above, this is an overrun of four doubles.

The offset should be:

```cpp
T* more_work = work + k*k;
```

The same correction is needed in both example files. After applying it and
rebuilding, the synthetic example completes normally and reports small angles
between the recovered and reference singular vectors.

Although Windows heap validation exposed the error as status `0xc0000374`, the
out-of-bounds write is cross-platform and may silently corrupt memory on other
systems.

Environment tested: Windows x64, MSVC 19.51, C++20, NMake generator,
LAPACK++/BLAS++ with standalone Intel oneMKL ILP64 sequential libraries.

### DIFF
--- a/RandBLAS/testing/comparison.hh
+++ b/RandBLAS/testing/comparison.hh
@@ -36,6 +36,16 @@
 #include <string>
 #include "RandBLAS.hh"
 
+// Compiler-specific spellings for the function signature printed in test
+// failure diagnostics. Keep this test utility out of the installed RandBLAS
+// target's compiler definitions.
+#if defined(_MSC_VER)
+    #define RANDBLAS_TEST_FUNCTION_SIGNATURE __FUNCSIG__
+#elif defined(__GNUC__) || defined(__clang__)
+    #define RANDBLAS_TEST_FUNCTION_SIGNATURE __PRETTY_FUNCTION__
+#else
+    #define RANDBLAS_TEST_FUNCTION_SIGNATURE __func__
+#endif
 
 namespace RandBLAS::testing {
 
--- a/RandBLAS/testing/comparison.hh
+++ b/RandBLAS/testing/comparison.hh
@@ -40,11 +40,11 @@
 // failure diagnostics. Keep this test utility out of the installed RandBLAS
 // target's compiler definitions.
 #if defined(_MSC_VER)
-    #define RANDBLAS_TEST_FUNCTION_SIGNATURE __FUNCSIG__
+    #define __RANDBLAS_PRETTY_FUNCTION__ __FUNCSIG__
 #elif defined(__GNUC__) || defined(__clang__)
-    #define RANDBLAS_TEST_FUNCTION_SIGNATURE __PRETTY_FUNCTION__
+    #define __RANDBLAS_PRETTY_FUNCTION__ __PRETTY_FUNCTION__
 #else
-    #define RANDBLAS_TEST_FUNCTION_SIGNATURE __func__
+    #define __RANDBLAS_PRETTY_FUNCTION__ __func__
 #endif
 
 namespace RandBLAS::testing {
--- a/examples/sparse-low-rank-approx/svd_matrixmarket.cc
+++ b/examples/sparse-low-rank-approx/svd_matrixmarket.cc
@@ -198,7 +198,7 @@ void qb_to_svd(int64_t m, int64_t n, int64_t k, T* Q, T* svals, int64_t ldq, T* 
 
     // update Q = Q U.
     bool allocate_more_work = extra_work_size < m*k;
-    T* more_work = (allocate_more_work) ? new T[m*k] : (work + k*(k+1)); 
+    T* more_work = (allocate_more_work) ? new T[m*k] : (work + k*k);
     lapack::lacpy(MatrixType::General, m, k, Q, ldq, more_work, m);
     blas::gemm(Layout::ColMajor, Op::NoTrans, Op::NoTrans, m, k, k, 1.0, more_work, m, U, k, 0.0, Q, ldq);
 
--- a/examples/sparse-low-rank-approx/svd_matrixmarket.cc
+++ b/examples/sparse-low-rank-approx/svd_matrixmarket.cc
@@ -32,7 +32,6 @@
 #include <lapack.hh>
 #include <omp.h>
 #include <stdio.h>
-#include <unistd.h>
 #include <iostream>
 #include <cmath>
 #include <time.h>
--- a/.github/actions/setup-randblas-deps-windows/action.yml
+++ b/.github/actions/setup-randblas-deps-windows/action.yml
@@ -8,6 +8,10 @@ inputs:
     description: "Build and install LAPACK++ for the examples workflow."
     required: false
     default: "false"
+  sanitize-address:
+    description: "Build GoogleTest with the MSVC AddressSanitizer enabled."
+    required: false
+    default: "false"
 
 outputs:
   blaspp-dir:
@@ -36,10 +40,18 @@ runs:
         key: windows-vcpkg-intel-mkl-2025.2.0-1-${{ hashFiles('.github/actions/setup-randblas-deps-windows/setup.ps1') }}
 
     - name: cache GoogleTest installation
+      if: inputs.sanitize-address != 'true'
       uses: actions/cache@v4
       with:
         path: ${{ github.workspace }}\..\windows-deps\googletest-install
         key: windows-msvc-googletest-1.17.0-1
+
+    - name: cache AddressSanitizer GoogleTest installation
+      if: inputs.sanitize-address == 'true'
+      uses: actions/cache@v4
+      with:
+        path: ${{ github.workspace }}\..\windows-deps\googletest-asan-install
+        key: windows-msvc-googletest-asan-1.17.0-1-${{ hashFiles('.github/actions/setup-randblas-deps-windows/setup.ps1') }}
 
     - name: cache BLAS++ installation
       uses: actions/cache@v4
@@ -65,6 +77,8 @@ runs:
       shell: pwsh
       run: |
         $installLapackpp = '${{ inputs.install-lapackpp }}' -eq 'true'
+        $sanitizeAddress = '${{ inputs.sanitize-address }}' -eq 'true'
         & "${env:GITHUB_ACTION_PATH}\setup.ps1" `
           -DependencyRoot "${env:GITHUB_WORKSPACE}\..\windows-deps" `
-          -InstallLapackpp:$installLapackpp
+          -InstallLapackpp:$installLapackpp `
+          -SanitizeAddress:$sanitizeAddress
--- a/.github/actions/setup-randblas-deps-windows/action.yml
+++ b/.github/actions/setup-randblas-deps-windows/action.yml
@@ -29,29 +29,6 @@ outputs:
 runs:
   using: composite
   steps:
-    - name: resolve dependency revisions
-      id: revisions
-      shell: pwsh
-      run: |
-        $ErrorActionPreference = "Stop"
-
-        function Get-RemoteHead([string] $url) {
-          $line = (& git ls-remote $url HEAD | Select-Object -First 1)
-          if ($LASTEXITCODE -ne 0 -or -not $line) {
-            throw "Could not resolve HEAD for $url"
-          }
-          return ($line -split "\s+")[0]
-        }
-
-        "blaspp=$(Get-RemoteHead 'https://github.com/icl-utk-edu/blaspp.git')" |
-          Out-File -FilePath $env:GITHUB_OUTPUT -Append -Encoding utf8
-        "random123=$(Get-RemoteHead 'https://github.com/DEShawResearch/Random123.git')" |
-          Out-File -FilePath $env:GITHUB_OUTPUT -Append -Encoding utf8
-        if ('${{ inputs.install-lapackpp }}' -eq 'true') {
-          "lapackpp=$(Get-RemoteHead 'https://github.com/icl-utk-edu/lapackpp.git')" |
-            Out-File -FilePath $env:GITHUB_OUTPUT -Append -Encoding utf8
-        }
-
     - name: cache vcpkg oneMKL installation
       uses: actions/cache@v4
       with:
@@ -68,20 +45,20 @@ runs:
       uses: actions/cache@v4
       with:
         path: ${{ github.workspace }}\..\windows-deps\blaspp-install
-        key: windows-msvc-blaspp-ilp64-sequential-1-${{ steps.revisions.outputs.blaspp }}-${{ hashFiles('.github/actions/setup-randblas-deps-windows/setup.ps1') }}
+        key: windows-msvc-blaspp-windows-portability-ilp64-sequential-3-${{ hashFiles('.github/actions/setup-randblas-deps-windows/setup.ps1') }}
 
     - name: cache Random123 installation
       uses: actions/cache@v4
       with:
         path: ${{ github.workspace }}\..\windows-deps\Random123-install
-        key: windows-random123-1-${{ steps.revisions.outputs.random123 }}
+        key: windows-random123-2-${{ hashFiles('.github/actions/setup-randblas-deps-windows/setup.ps1') }}
 
     - name: cache LAPACK++ installation
       if: inputs.install-lapackpp == 'true'
       uses: actions/cache@v4
       with:
         path: ${{ github.workspace }}\..\windows-deps\lapackpp-install
-        key: windows-msvc-lapackpp-ilp64-sequential-1-${{ steps.revisions.outputs.blaspp }}-${{ steps.revisions.outputs.lapackpp }}-${{ hashFiles('.github/actions/setup-randblas-deps-windows/setup.ps1') }}
+        key: windows-msvc-lapackpp-msvc-compatibility-blaspp-windows-portability-ilp64-sequential-3-${{ hashFiles('.github/actions/setup-randblas-deps-windows/setup.ps1') }}
 
     - name: build missing Windows dependencies
       id: setup
--- a/.github/actions/setup-randblas-deps-windows/action.yml
+++ b/.github/actions/setup-randblas-deps-windows/action.yml
@@ -1,0 +1,93 @@
+name: setup-randblas-deps-windows
+description: >
+  Provisions oneMKL through vcpkg and builds the native-MSVC dependencies used
+  by RandBLAS's Windows CI jobs.
+
+inputs:
+  install-lapackpp:
+    description: "Build and install LAPACK++ for the examples workflow."
+    required: false
+    default: "false"
+
+outputs:
+  blaspp-dir:
+    description: "Directory containing blasppConfig.cmake."
+    value: ${{ steps.setup.outputs.blaspp-dir }}
+  random123-dir:
+    description: "Directory containing the installed Random123 headers."
+    value: ${{ steps.setup.outputs.random123-dir }}
+  googletest-prefix:
+    description: "GoogleTest installation prefix."
+    value: ${{ steps.setup.outputs.googletest-prefix }}
+  lapackpp-dir:
+    description: "Directory containing lapackppConfig.cmake, when requested."
+    value: ${{ steps.setup.outputs.lapackpp-dir }}
+  mkl-root:
+    description: "Root of the vcpkg oneMKL installation."
+    value: ${{ steps.setup.outputs.mkl-root }}
+
+runs:
+  using: composite
+  steps:
+    - name: resolve dependency revisions
+      id: revisions
+      shell: pwsh
+      run: |
+        $ErrorActionPreference = "Stop"
+
+        function Get-RemoteHead([string] $url) {
+          $line = (& git ls-remote $url HEAD | Select-Object -First 1)
+          if ($LASTEXITCODE -ne 0 -or -not $line) {
+            throw "Could not resolve HEAD for $url"
+          }
+          return ($line -split "\s+")[0]
+        }
+
+        "blaspp=$(Get-RemoteHead 'https://github.com/icl-utk-edu/blaspp.git')" |
+          Out-File -FilePath $env:GITHUB_OUTPUT -Append -Encoding utf8
+        "random123=$(Get-RemoteHead 'https://github.com/DEShawResearch/Random123.git')" |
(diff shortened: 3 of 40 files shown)


rileyjmurray:
Thanks for this contribution, @RaphaelArkadyMeyerNYU! Would you like me to give a review the next chance I get, or are there things you'd like to change first?

RaphaelArkadyMeyerNYU:
Eh, I'm not sure.

It's unclear to me what the standard for "supporting windows" really is.

Seems like there's certainly some low-hanging fruit (removing the unused linux-only header file in some of the example codes).

But beyond that, seems like we should think about the right way to handle the Random123 and Blas++ little incompatibilities and the openMP support. Imo, all of those need some level of _discussion_.

rileyjmurray:
Follow-up from our discussions last night: the standard is just "we have a Windows CI build that works." I suggest only using MSVC in the Windows build. If someone uses gcc or clang on Windows it'll probably be easy for them to resolve small problems / quirks they encounter. BLAS should use 64-bit integers.

RaphaelArkadyMeyerNYU:
Hey @rileyjmurray , can you allow this to build via the github action again (I don't think I can enable that to happen, if I understand that correctly?).

I'd like to make sure my suite of further CMake changes and a few minor edits to the examples didn't break any of the existing checks before making a more self-contained windows setup system.

RaphaelArkadyMeyerNYU:
Hi @rileyjmurray. I worked at further and further things to make the installing process on windows not be super janky (incl: a branch [here](https://github.com/RaphaelArkadyMeyerNYU/blaspp/tree/windows-portability) for updating Blas++ and one [here](https://github.com/RaphaelArkadyMeyerNYU/lapackpp/tree/msvc-compatibility) for Lapack++).

Ended up being a lot of edits, but nothing to core RandBLAS.

There's updated information in install.md, as well as a new suite of github actions.

Not sure there's anything else I can/should do before we look at merging this into main. If so, let me know!

examples/sparse-low-rank-approx/svd_matrixmarket.cc
rileyjmurray, line 1: I take it that shifting by k*(k+1) was too far, leading to a segfault, and the correct shift is just k*k?

RandBLAS/testing/comparison.hh
rileyjmurray, line 1: Clever!

rileyjmurray (commented):
This is incredible! I was afraid when I saw +1000 lines and 40 files changed. After review I now see that the vast majority of changed lines are CI plus the installation guide, while the vast majority of changed files just have substitutions of the `__PRETTY_FUNCTION__`macro.

I have one request: can you rename `RANDBLAS_TEST_FUNCTION_SIGNATURE` to `__RANDBLAS_PRETTY_FUNCTION__`?

I have one question too, about a change in pointer offset value. Please respond to that in the comment thread when you get a chance.

rileyjmurray:
Ping @RaphaelArkadyMeyerNYU. This is basically ready to merge. Just a small request and a question!

rileyjmurray:
Looks like that rename attempt didn't work

rileyjmurray (approved):

rileyjmurray:
@RaphaelArkadyMeyerNYU can you edit the PR's comment to reflect the PR's current state? I think some of the issues have been handled in your revisions. For example, `D__PRETTY_FUNCTION__=__FUNCSIG__` is no longer needed, since RandBLAS' uses of `__PRETTY_FUNCTION__` have been replaced by a portable RandBLAS-defined macro (`__RANDBLAS_PRETTY_FUNCTION__`) that does the same thing.

RaphaelArkadyMeyerNYU:
@rileyjmurray I replaced it with the LLM-updated summary of all the points. Imo it's pretty much all fixed, modulo upstream issues and a minor question about which OpenMP versions should be allowed on MSVC (namely, the pull forces that OpenMP is run with the "experimental" flag on MSVC; this is because a certain parallelization command RandBLAS uses is currently only available in experimental MSVC OpenMP).

Does this look good / work for an updated PR comment? Good to merge?

rileyjmurray:
LET'S GOOOOOO!!!

PR #183, opened 2026-08-05
INSTALL.md: clarify the RandLAPACK-submodule relationship

mmelnich:
States the submodule rules where RandBLAS developers will read them: develop RandBLAS in its own clone, never inside RandLAPACK's submodule checkout; never point RandLAPACK at a RandBLAS working copy; a standalone RandBLAS install does not affect RandLAPACK. Points package maintainers at RandLAPACK's new RandLAPACK_EXTERNAL_RandBLAS option, which admits an external RandBLAS only when built from exactly the pinned commit. Companion to the RandLAPACK-side PR.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -75,6 +75,18 @@ for more info.
 
 ## 3. Building and installing RandBLAS
 
+**A note if you arrived here from RandLAPACK.** RandLAPACK vendors RandBLAS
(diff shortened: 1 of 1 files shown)


mmelnich:
@rileyjmurray I am working on tuning the connection between RandBLAS and RandLAPACK & figured I'd add an explicit statement here.

rileyjmurray (approved):

PR #184, opened 2026-08-06
MSVC OpenMP: switch to the /openmp:llvm runtime

mmelnich:
Follow-up to the Windows portability work in #179.

**What #179 did:** made RandBLAS build natively with MSVC, including the CMake logic that selects an MSVC OpenMP mode. It chose `/openmp:experimental`, the minimum needed for RandBLAS itself: plain `/openmp` cannot compile the `omp simd` directive in the sparse kernels.

**What this PR changes:** the selected mode becomes `/openmp:llvm`. The experimental mode is still OpenMP 2.0 underneath, which only allows signed 32-bit loop indices and has no `collapse` clause. RandLAPACK parallelizes loops with `int64_t` indices and uses `collapse(2)`, so with the experimental mode pinned here, RandLAPACK's native Windows builds (BallisticLA/RandLAPACK#154) have to disable OpenMP entirely. The LLVM mode supports 64-bit indices, `collapse`, and everything the experimental mode provided, including `omp simd` — so RandBLAS keeps compiling as before and downstream consumers stop being blocked.

**Also fixed along the way:** the re-stamp of `OpenMP::OpenMP_CXX` after `find_package` hardcoded the mode string, so a user passing `-DOpenMP_CXX_FLAGS=...` at configure time was silently overridden. The requested flags are now captured before the find and re-applied afterward, in both the build tree (`CMake/OpenMP.cmake`) and the installed package config (`RandBLASConfig.cmake.in`).

**Validation:** the Windows CI matrix from #179 (openmp-release leg, 459 tests) is the gate — it exercises the sparse kernels' `omp simd` under the new mode. Serial and ASan legs are unaffected.

**Downstream plan:** once this merges, RandLAPACK bumps its submodule pin and adds an OpenMP leg to its own Windows CI.

### DIFF
--- a/CMake/OpenMP.cmake
+++ b/CMake/OpenMP.cmake
@@ -1,21 +1,27 @@
 message(STATUS "Checking for OpenMP ... ")
 
-# MSVC's ordinary /openmp mode does not support the omp simd directive used by
-# the sparse kernels. Recent MSVC versions provide it through
-# /openmp:experimental. Allow callers to override this setting explicitly.
+# MSVC's classic /openmp mode implements OpenMP 2.0 only: it rejects the
+# omp simd directive used by the sparse kernels, 64-bit loop indices, and
+# the collapse clause used by downstream consumers such as RandLAPACK. The
+# /openmp:llvm runtime supports all of these (and subsumes what
+# /openmp:experimental offered). Callers can still override the mode
+# explicitly with -DOpenMP_CXX_FLAGS=... at configure time.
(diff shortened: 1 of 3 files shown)


mmelnich:
@rileyjmurray Need this approved to finish my RandLAPACK PR

mmelnich:
@rileyjmurray All checks pass now.

rileyjmurray (approved):

